Issue #3, opened 2018-06-27
Support for optional dataclass attributes


ElmoVanKielmo:
Good job, but `_decode_dataclass(cls, kvs)` should be aware of optional attributes.  
Consider the following code:
```python
import typing

from dataclasses import dataclass
from dataclasses_json import DataClassJsonMixin


@dataclass
class Foo(DataClassJsonMixin):
    bar: float
    baz: typing.Optional[float]


Foo.from_json('{"bar": 7.45}')
```
This throws `KeyError` despite `baz` being optional.

ElmoVanKielmo:
Of course I meant `to_json()` and not `to_dict()` in description of my second commit ;)

lidatong:
Thanks for raising the issue and the corresponding PR!

I've given this some thought and I actually don't agree that `{"bar": 7.45}` should be a valid json representation of the `Foo` class in your example.

When considering the distinction between:
{"bar": 7.45}
vs.
{"bar": 7.45, "baz": None}

The main problem I see with making the former the json representation of a `Foo` is it would lead to equivalent json representations of two structurally dissimilar types:

```python
class Foo:
    bar: float

class Foo:
    bar: float
    baz: Optional[float]

# these two classes end up having the same json repr
```

However, I'm in favor of adding an option to `from_json` to "infer missing keys as none" (which would be similar to your pr). Technically, this violates a (potential) invariant that `from_json` and `to_json` are inverses, which as nice that may be to provide, supplementing the API to support this usecase seems the more practical approach.

ElmoVanKielmo:
I completely agree.

lidatong:
really sorry for the long delay. `infer_missing` feature provided as a kwarg to `from_json` in https://github.com/lidatong/dataclasses-json/releases/tag/v0.0.7 along with fixes to make this lib work in 3.7

PR #4, opened 2018-06-27
Fix for https://github.com/lidatong/dataclasses-json/issues/3

ElmoVanKielmo:
This pull request enables possibility to decode json if optional dataclass fields are missing, as well as encoding dataclass to json without optional fields if their values are `None`.
Encoding behavior is controlled by `skip_missing_optionals` keyword argument for `DataClassJsonMixin.to_json()` which defaults to `False` for backwards compatibility.


lidatong:
Addressed in #3

PR #14, opened 2018-09-11
Check str in __args__ for type

hasanatkazmi:
These changes fix issues 2 and 3.
Too small to write any description. Issue 2 and 3 describe the problem.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -41,7 +41,7 @@ def _decode_dataclass(cls, kvs, infer_missing):
                                                         field_value,
                                                         infer_missing)
 
-        elif _is_supported_generic(field.type) and field.type != str:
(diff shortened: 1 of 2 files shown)


dataclasses_json/core.py
```diff
@@ -41,7 +41,7 @@ def _decode_dataclass(cls, kvs, infer_missing):
                                                         field_value,
                                                         infer_missing)
 
-        elif _is_supported_generic(field.type) and field.type != str:
+        elif _is_supported_generic(field.type) and field.type != str and field.type.__args__[0] != str:
```
lidatong: Thanks for putting in this fix, but I've found the issue actually stems from `str` passing the `_is_supported_generic` check. See #16

lidatong:
Fix in #16

PR #41, opened 2018-11-14
Add support for passing the MM fields directly

canassa:
This adds support for passing the Marshmallow fields directly to the dataclass.

This is done via the dataclass field metadata: e.g.:

```python
@dataclass
class Car(DataClassJsonMixin):
    license_number: str = field(metadata={'mm': fields.String(required=False)})
```

The metadata field takes priority over the dataclass type.

I probably forgot a bunch of edge cases, so please review it care 😁

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -85,11 +85,20 @@ class DataClassJsonMixin(abc.ABC):
             return _decode_dataclass(cls, kvs, infer_missing=partial)
 
         fields_ = fields(cls)
-        nested_fields = mm._make_nested_fields(fields_,
+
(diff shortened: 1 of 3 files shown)


lidatong:
Hi, thanks for raising this use case and taking the time to provide the accompanying solution & PR!

Coincidentally someone else just raised the issue that would require overriding default behavior.

The [latest release](https://github.com/lidatong/dataclasses-json/releases) has this feature and also  updated [docs](https://github.com/lidatong/dataclasses-json#overriding) on how to do it. As a bonus, you can also override the `encoder`/`decoder` separately from the `mm_field`. Give it a shot!

Issue #40, opened 2018-11-14
0.1.0 datetime handling does not conform to iso 8601


korobkov-mindbox:
In JSON serialization [it is generally considered standard](https://stackoverflow.com/questions/10286204/the-right-json-date-format) to use string-like datetime format. For example, .NET Newtonsoft.JSON serializes datetime values as strings: `{"Now":"2018-11-14T15:47:11.881469+03:00"}`, same as javascript's JSON library.

Prior to 0.1.0 release it was possible to write our own __post_init__ method to decode such datetime fields:
```
@dataclass_json
@dataclass(frozen=True)
class ExampleObject:
    SomeDate: datetime

    def __post_init__(self):
        current_value = self.__dict__['SomeDate']
        if isinstance(current_value, str):
            self.__dict__['SomeDate'] = dateutil.parser.parse(current_value)
```

Release 0.1.0 breaks this behaviour and generally makes it impossible to work with commonly-accepted string-like rfc/iso datetime formats by always forcing datetime representation to be int-like due to [this code](https://github.com/lidatong/dataclasses-json/commit/36aca833d070d10c36f6743e1d8d38f67f46bb1b#diff-c9f66dd085ddaa057e97df6fdb0eb6c3R92).
```
TypeError : an integer is required (got type str)
at
...
File "api.py", line 65, in from_json
return _decode_dataclass(cls, init_kwargs, infer_missing)
File "core.py", line 69, in _decode_dataclass
infer_missing)
File "core.py", line 84, in _decode_dataclass
dt = datetime.fromtimestamp(field_value, tz=tz)
```

As of now we are forced to use 0.0.25 release, however it would be of great help if we could rely on dataclasses-json to natively handle datetime representations without our own hacks.

Btw, using string-like representation also has inherent benefit of allowing you to store timezone info in the serialized value, ensuring that encoding->decoding->encoding operation is strictly identity (note that d->e->d still will not be strict identity because serialization format may change, but that is okay as you still keep all timezone-related information)

lidatong:
Hi, thanks for raising this use case!

I agree that ISO is a standard format but would say UNIX time is also a standard format when working with datetime. There were tradeoffs I considered when picking one or the other, and I decided to go with UNIX time via `timestamp` primarily because I think it's simpler (especially in that it doesn't require the client of the JSON to have an ISO parser). 

[...]

PR #43, opened 2018-11-17
Add support for field overrides

lidatong:


lidatong:
Addresses #40 and #41 

Supersedes #41 (thanks @canassa!)

PR #51, opened 2018-12-04
Add support for Enum

HappyTreeBeard:
This change allows the package to convert JSON to ```enum.Enum``` utilizing the type hint of the dataclass field. The *.value property of an Enum is used when encoding JSON. Unit tests to validate these changes are included and follow the same structure @lidatong has been using in existing code.. 

``` python
from enum import Enum


class MyEnum(Enum):
    STR1 = "str1"
    INT1 = 1


@dataclass_json
@dataclass
class DataWithEnum:
    name: str
    my_enum: MyEnum


x = DataWithEnum.from_json('{"name": "name2", "my_enum": "str1"}')
# DataWithEnum(name='name2', my_enum=<MyEnum.STR1: 'str1'>)
assert x.my_enum == MyEnum.STR1
```

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -1,6 +1,8 @@
 import copy
 import json
 import warnings
+from enum import Enum
+
 from dataclasses import MISSING, _is_dataclass_instance, fields, is_dataclass
 from datetime import datetime, timezone
 from typing import Collection, Mapping, Union
@@ -26,6 +28,8 @@ class _ExtendedEncoder(json.JSONEncoder):
             result = o.timestamp()
         elif _isinstance_safe(o, UUID):
             result = str(o)
+        elif _isinstance_safe(o, Enum):
+            result = o.value
         else:
             result = json.JSONEncoder.default(self, o)
         return result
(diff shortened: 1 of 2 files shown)


HappyTreeBeard:
I realized these changes would only allow Enum support when using the ```to_json``` and ```from_json``` functions. These changes would not support Enum if using the Marshmallow schema object.

Marshmallow does not support an Enum field but a separate [marshmallow-enum](https://github.com/justanr/marshmallow_enum) package could be used.

HappyTreeBeard:
@lidatong, what do you think about this change and do you think it could be merged?

lidatong:
Hi, I'm sorry I forgot to get back to you on this. I like this idea, and I think it would be great to add it to the library!

However, I think this feature isn't quite done:

1. As you pointed out, given this PR, there would be an API inconsistency when using marshmallow functionality, e.g. `.schema().load`. However, it is certainly possible to make this library generate marshmallow schemas for enums, perhaps with the help of the marshmallow_enum library.

2. I also want to make sure this handles non-primitive Enum values (e.g. dataclasses inside `.value`).

Rather than merging into master, I'm going to merge this into a feature branch for a release. I plan to refactor some of this library's core (mainly the current handling of additional types is rather hacky / buggy...), which will also conveniently make this feature easier to finish implementing. I've also added you as a contributor, and would greatly appreciate if you continued to work on the feature branch directly in the main repo (if you like). Also, thanks for your contribution I appreciate it a lot!

lidatong:
Also, I want to very much apologize again for the delay. I've been quite busy, but am aiming to dive back into working on the library and plan to be working on this across the next few days.

Issue #52, opened 2018-12-07
datetime.datetime.fromisoformat Compatibility


HappyTreeBeard:
The function ```datetime.datetime.fromisoformat(str)``` is [new to Python 3.7](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) and ```setup.py``` states ``` python_requires=">=3.6"``` The [backports-datetime-fromisoformat](https://pypi.org/project/backports-datetime-fromisoformat/) package on pypi should be added to the package requirements if Python 3.6 is to be supported.

 It looks like the ```fromisoformat``` dependency was added to resolve the ["0.1.0 datetime handling does not conform to iso 8601"](https://github.com/lidatong/dataclasses-json/issues/40) ticket.

lidatong:
Please see explanation in #53

PR #53, opened 2018-12-30
Backport fromisoformat

HappyTreeBeard:
These changes allow unit tests to run in Python 3.6. As mentioned in  #52, ```datetime.fromisoformat``` was implemented in Python 3.7. 

This does not fix the issue where ```dataclasses_json.mm._IsoField._deserialize``` calls ```datetime.fromisoformat```. Why was the _IsoField class created as a protected class and not a normal publicly facing class?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,5 +23,10 @@ setup(
     extras_require={
         "dev": ["pytest", "ipython", "mypy", "hypothesis"]
     },
+    tests_require=[
+        "pytest",
+        "hypothesis",
+        "backports-datetime-fromisoformat;python_version<'3.7'",
+    ],
     include_package_data=True
 )
--- a/tests/entities.py
+++ b/tests/entities.py
@@ -17,6 +17,13 @@ from marshmallow import fields
 from dataclasses_json import DataClassJsonMixin, dataclass_json
 from dataclasses_json.mm import _IsoField
 
+if not hasattr(datetime, 'fromisoformat'):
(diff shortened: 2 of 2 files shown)


lidatong:
Hi, to provide some more context (and I believe this will address #52 as well), I purposely made the class private to indicate it's not intentionally part of the public API, and could be removed unexpectedly.

This is because ideally marshmallow already serializes/deserializes to ISO format by default: there just is an outstanding bug where it doesn't work properly (it returns `naive` objects when it's documented to return `aware`). Ref: https://github.com/marshmallow-code/marshmallow/issues/955

So I created this class in `mm.py` as a temporary fix (with the intention of removing it once marshmallow fixed the bug). This probably led to confusion, and I'm sorry about that.

lidatong:
Thank you for catching there is a 3.6 compatibility issue in the tests!

To address the `isoformat` thing:

I didn't intend to actually allow my implementation of a datetime serde using ISO format to be publicly exposed: rather I intended the user to implement it themselves via an override (and it's up to them whether they want to do that with 3.7's `datetime.fromisoformat` or `python-dateutil`, etc.).

The idea is doing that would be "just like any other field type override" by the user. I thought about the trade-offs of providing a special flag (or something along those lines) in the API specifically for datetime serde format, but decided it would be better to just give the user the power to decide themselves via override (and it's also more general that way).

lidatong:
Fixed in #66 by versioning tests to only run in 3.7

Issue #45, opened 2018-11-20
Troubble with boolean fields


moreklatring:
I want to have some fields excluded, so I am using call  b) below. But it throws an exception for the boolean field. Is this a bug? 

`@dataclass_json
@dataclass
class Foo:
    bar: bool = False
    zoo: str = "Exclude me"

d = Foo()

# a)
print(d.to_json())

# b)
print(Foo.schema(exclude=["zoo"]).dumps(d))`

HappyTreeBeard:
The problem stems from the function ```dataclasses_json.mm._make_default_field(type_, default, cls)``` when initializing a marshmallow class.  https://github.com/lidatong/dataclasses-json/blob/0f48220ee711e94cfdb8aae7143702ce47a0e53e/dataclasses_json/mm.py#L112

Why is ```cls``` passed as a positional argument to the marshmallow constructor? I believe this is a bug and I believe this bug has been overlooked because it's only an issue for ```fields.Bool``` which is the only constructor that accepts positional arguments. 

Note that the ```cls``` should be passed as a positional argument if ```fields.List``` is used as seen on line 111.

https://github.com/lidatong/dataclasses-json/blob/0f48220ee711e94cfdb8aae7143702ce47a0e53e/dataclasses_json/mm.py#L104-L112

This makes me think it was a copy and paste error.

lidatong:
Hi, yes, thanks to @HappyTreeBeard the issue is resolved. Indeed, the bug looks like it stems from a sloppy typo (possibly due to copy paste 😅

PR #54, opened 2018-12-30
Bug fix for default bool when using schema

HappyTreeBeard:
This change fixes the issues discussed in #45 where default value boolean fields using marshmallow schema.  Unit tests were added to confirm the problem has been fixed.


lidatong:
Hi, thank you so much for the fix(es) and I'm sorry for the delay in me reviewing this.

PR #55, opened 2018-12-30
Fix broken warning message

HappyTreeBeard:
Original bug most likely created when refactoring. ```cls``` variable changed to ```dataclass_json_cls```.


lidatong:
Thanks for the fix!

PR #64, opened 2019-02-14
Feature add decimal support

exaroth:
This adds basic support for decimal type which will by default raise an error when trying to deserialize value back to json (at least for nested dataclasses). Input value should likely be string to maintain precision however it should also work fine with floats/ints in input json.


lidatong:
Hey there, thanks for putting in this PR, I appreciate it a lot.

This is actually a problematic way to serialize decimals though, as calling `float(...)` results in a loss of precision (which is probably not we want, since one primary use-case of `Decimal` is numeric precision). The best way to address this is probably to serialize the decimal as a string

lidatong:
Closing as superseded by #86

Issue #65, opened 2019-02-15
dump/load gives Invalid input type


neggert:
Using `schema().load()` does not seem to work for classes that contain lists. I can reproduce this even with the Boss/Minion example from the documentation

```{python}
@dataclass_json
@dataclass(frozen=True)
class Minion:
    name: str


@dataclass_json
@dataclass(frozen=True)
class Boss:
    minions: List[Minion]

boss = Boss([Minion('evil minion'), Minion('very evil minion')])
boss_dict = Boss.schema().dump(boss)
Boss.schema().load(boss_dict)
```

```
marshmallow.exceptions.ValidationError: {'_schema': ['Invalid input type.']}
```

For what it's worth, going all the way to JSON and back seems to work fine:

```
boss_json = boss.to_json()
Boss.from_json(boss_json)
```

nuklea:
Same issue with 0.2.1

nuklea:
```python
from dataclasses import dataclass
from typing import List

from dataclasses_json import DataClassJsonMixin


@dataclass
class Entity(DataClassJsonMixin):
    id: int


@dataclass
class Container(DataClassJsonMixin):
    entities: List[Entity]


@dataclass
class Data(DataClassJsonMixin):
    container: Container


schema = Data.schema()
schema.load({'container': {'entities': [{'id': 1}]}})
```

raises `marshmallow.exceptions.ValidationError: {'container': {'entities': {'_schema': ['Invalid input type.']}}}`

nuklea:
But works fine with 0.1.0

lidatong:
Thanks for reporting this. Fixed in #66

Issue #58, opened 2019-01-22
Support for dataclass fields of type Optional[List[X]]


kylegentle:
### Summary
The library currently breaks for dataclasses with fields of type Optional[List[X]], returning the following error:
```Traceback (most recent call last):
  File "dcj_bug/ex.py", line 66, in <module>
    Adult.schema().loads(test3, many=True)  # Fail
  File "<home_path>/.cache/pypoetry/virtualenvs/dcj-bug-py3.7/lib/python3.7/site-packages/dataclasses_json/api.py", line 100, in schema
    infer_missing)
  File "<home_path>/.cache/pypoetry/virtualenvs/dcj-bug-py3.7/lib/python3.7/site-packages/dataclasses_json/mm.py", line 86, in _make_default_fields
    cls)
  File "<home_path>/.cache/pypoetry/virtualenvs/dcj-bug-py3.7/lib/python3.7/site-packages/dataclasses_json/mm.py", line 103, in _make_default_field
    cons = _type_to_cons[cons_type]
KeyError: typing.List[__main__.Child]
```

I've attached some code below that demonstrates the issue.

Please let me know if there's any information or assistance I can provide; I dug into the code a bit in mm.py where it's breaking, but wasn't able to figure out what the correct logic should be in mm._make_default_field.

### Example
```python
import json

from dataclasses import dataclass
from typing import Optional, List

from dataclasses_json import DataClassJsonMixin


@dataclass
class Child(DataClassJsonMixin):
    name: str


@dataclass
class Adult(DataClassJsonMixin):
    name: str
    children: Optional[List[Child]] = None


test1 = """
{
    "name": "Foo"

}
""".strip()

test2 = """
{
    "name": "Bar",
    "children": [
        {
            "name": "Baz"
        },
        {
            "name": "Bat"
        }
    ]
}
""".strip()

test3 = """
[
    {
        "name": "Foo"
    },
    {
        "name": "Bar",
        "children": [
            {
                "name": "Baz"
            },
            {
                "name": "Bat"
            }
        ]
    }
]
""".strip()

test4 = json.loads(test3)

Adult.from_json(test1)  # OK
Adult.from_json(test2)  # OK

# Adult.from_json(test3)  # Breaks
Adult.schema().loads(test3, many=True)  # Breaks
# Adult.schema().load(test4, many=True)  # Breaks
```

kylegentle:
[...]

PR #66, opened 2019-02-19
Fix schema problems

### LINKED ISSUE #57: Invalid input type
Hello,

The title shows the error i've been having with this library. I've put together a scenario which reproduces the problem:

```
from typing import List

@dataclass_json
@dataclass
class Relative:
    name: str

@dataclass_json
@dataclass
class Person:
    name: str
    relatives: List[Relative]


people_json = '[{"name": "lidatong", "relatives": [{"name": "relative 1"}, {"name": "relative 2"}]}]'
Person.schema().loads(people_json, many=True)

```

The stacktrace i get is:
```
  File "main.py", line 20, in <module>
    Person.schema().loads(people_json, many=True)
  File "/home/rakan/.pyenv/versions/proj/lib/python3.7/site-packages/marshmallow/schema.py", line 724, in loads
    return self.load(data, many=many, partial=partial, unknown=unknown)
  File "/home/rakan/.pyenv/versions/proj/lib/python3.7/site-packages/marshmallow/schema.py", line 695, in load
    postprocess=True,
  File "/home/rakan/.pyenv/versions/sharq-urgent/lib/python3.7/site-packages/marshmallow/schema.py", line 857, in _do_load
    raise exc
marshmallow.exceptions.ValidationError: {0: {'relatives': {'_schema': ['Invalid input type.']}}}
```

### LINKED ISSUE #60: schema().dumps not working properly with field initializers
Consider the following dataclass:
```python
from typing import List

@dataclass_json
@dataclass
class Bar:
    foos: List[str] = field(default_factory=list)
```

using `Bar.schema().dumps(Bar())` gives me the following error:

```
  File "dataclass_test.py", line 20, in <module>
    Bar.schema().dumps(Bar())
  File "~/.venv/venv/lib/python3.6/site-packages/dataclasses_json/api.py", line 100, in schema
    infer_missing)
  File "~/.venv/venv/lib/python3.6/site-packages/dataclasses_json/mm.py", line 90, in _make_default_fields
    cls)
  File "~/.venv/venv/lib/python3.6/site-packages/dataclasses_json/mm.py", line 105, in _make_default_field
    type_arg = type_.___args__[0]
AttributeError: type object 'List' has no attribute '___args__'

Process finished with exit code 1
```

However, it works when the type of `foos` is a `List` of another `dataclass_json`, like this:
```python
@dataclass_json
@dataclass
class Foo:
    one: str


@dataclass_json
@dataclass
class Bar:
    foos: List[Foo] = field(default_factory=list)


if __name__ == "__main__":
    print(Bar.schema().dumps(Bar()))
```

So something seems to be off with types and field initializers

I am using `dataclasses-json==0.2.1`

Peilonrayz:
In light of #68 I have updated the code to fix #57, #58, #60, #65, #67 and #68. I have added relevant tests and ensured they pass in Python 3.6 and 3.7. (I've changed the code to allow skipping of the datetime tests.)

From this I have moved all schema code into `mm` so that the marshmallow code is contained. And this is basically just a re-write of `mm`. I've updated `api` to accommodate these changes.

Otherwise only `core` has been changed. This is because marshmallow now loads schema data into the relevant model. (This can be tested against with code in #58). I'm unsure how you'd want the data to be handled, and so I chose to return early so I don't mangle with `core` too. **please verify this is the intended behavior**.

This should keep all the previous logic around options, most are in `schema` such as specifying a default, and handling `Optional`. The other special piece of logic is when creating a DC, I kept the `field_many` option. Otherwise `build_type` now builds all the types you'd previously specified, and a couple more, in a recursive form, defaulting to `fields.Field` and a warning.

---

I would like to request that we change how you handle the `dataclasses_json.mm_field` so that if the value is `None` we generate one by default. This isn't something you did previously, and so I didn't change this.

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -4,8 +4,6 @@ from dataclasses import fields
 from datetime import datetime
 from typing import Any, Callable, List, Optional, Tuple, TypeVar, Union
 
-from marshmallow import Schema, post_load
-
 from dataclasses_json import mm
 from dataclasses_json.core import (_ExtendedEncoder, _asdict, _decode_dataclass,
                                    _overrides, _issubclass_safe,
@@ -78,50 +76,15 @@ class DataClassJsonMixin(abc.ABC):
                dump_only=(),
                partial=False,
                unknown=None):
-        Meta = type('Meta',
-                    (),
-                    {'fields': tuple(field.name for field in fields(cls))})
-
-        @post_load
-        def make_instance(self, kvs):
-            return _decode_dataclass(cls, kvs, partial)
-
-        overriden_fields = {k: v.mm_field
-                            for k, v in _overrides(cls).items()}
-        generated_fields = {field for field in fields(cls)
-                            if field.name not in overriden_fields}
-        nested_fields = mm._make_nested_fields(generated_fields,
-                                               DataClassJsonMixin,
-                                               infer_missing)
-        primitive_fields = [field for field in generated_fields
-                            if field.name not in nested_fields]
-        default_fields = mm._make_default_fields(primitive_fields,
-                                                 cls,
-                                                 infer_missing)
-
-        # datetime codec is different from what's specified by `fields.DateTime`
-        # so need to override what is inferred by `class Meta`
-        datetime_fields = {field.name: mm._TimestampField(cls)
-                           for field in primitive_fields
-                           if _issubclass_safe(field.type, datetime)
-                           and field.name not in default_fields}
-
-        DataClassSchema = type(f'{cls.__name__.capitalize()}Schema',
-                               (Schema,),
-                               {'Meta': Meta,
-                                f'make_{cls.__name__.lower()}': make_instance,
-                                **overriden_fields,
-                                **nested_fields,
-                                **default_fields,
-                                **datetime_fields})
-        return DataClassSchema(only=only,
-                               exclude=exclude,
-                               many=many,
-                               context=context,
-                               load_only=load_only,
-                               dump_only=dump_only,
-                               partial=partial,
-                               unknown=unknown)
+        Schema = mm.build_schema(cls, DataClassJsonMixin, infer_missing, partial)
+        return Schema(only=only,
+                      exclude=exclude,
+                      many=many,
+                      context=context,
+                      load_only=load_only,
+                      dump_only=dump_only,
+                      partial=partial,
+                      unknown=unknown)
 
 
 def dataclass_json(cls):
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -1,13 +1,10 @@
 import abc
 import json
-from dataclasses import fields
-from datetime import datetime
 from typing import Any, Callable, List, Optional, Tuple, TypeVar, Union
 
-from dataclasses_json import mm
+from dataclasses_json.mm import build_schema
 from dataclasses_json.core import (_ExtendedEncoder, _asdict, _decode_dataclass,
-                                   _overrides, _issubclass_safe,
-                                   _override)
+                                   _overrides, _override)
 
 A = TypeVar('A')
 B = TypeVar('B')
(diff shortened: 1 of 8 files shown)


dsvensson:
This fixes cases like this, hooray!

```
class Bar:
 b: str

class Foo:
  a: List[Bar]
```

But if Bar contains an Enum-field, that fails with:
```
  File ".../venv/lib/python3.7/site-packages/dataclasses_json/api.py", line 76, in schema
    Schema = build_schema(cls, DataClassJsonMixin, infer_missing, partial)
  File ".../venv/lib/python3.7/site-packages/dataclasses_json/mm.py", line 113, in build_schema
    schema_ = schema(cls, mixin, infer_missing)
  File ".../venv/lib/python3.7/site-packages/dataclasses_json/mm.py", line 98, in schema
    t = build_type(type_, options, mixin, field, cls)
  File ".../venv/lib/python3.7/site-packages/dataclasses_json/mm.py", line 73, in build_type
    return inner(type_, options)
  File ".../venv/lib/python3.7/site-packages/dataclasses_json/mm.py", line 72, in inner
    return field.Field(**options)
AttributeError: 'Field' object has no attribute 'Field'
```

Peilonrayz:
@dsvensson Looks like an easy fix, I used `field` rather than `fields` [here](https://github.com/lidatong/dataclasses-json/pull/66/files#diff-17d6f2f251ae64dcc26ae93bf8e23974R72).

Could you provide the code that made this error?

dsvensson:
@Peilonrayz Oh that was easy. It was indeed just that typo. Too much redacting, but I can provide a test case later today. There's also another test case that ought to be added, for list of classes where one field contains a list of classes. That didn't work before this branch. #57 . More or less extending the last decoder test in the test...nestedsomething..py.

Peilonrayz:
@dsvensson Yeah I had done that with another `fields.Fields`, thought I changed all of them, after that tho... :(

I've added the test you mention already - [test][0] & [object][1]. Note, you can reproduce this without the outer most datatype being a list. It's just that the library only allows decoding lists via `.schema`, otherwise `from_json` and other code doesn't make use the schema code.

  [0]: https://github.com/lidatong/dataclasses-json/pull/66/commits/463ea01090b0d6571bb56c403f09a994e724a39c#diff-e1bc6e780c4b78fd3a4a85bf7b8ae0acR18
  [1]: https://github.com/lidatong/dataclasses-json/pull/66/commits/463ea01090b0d6571bb56c403f09a994e724a39c#diff-753922789de93c054356e2f67368b4d9R219

lidatong:
> In light of #68 I have updated the code to fix #57, #58, #60, #65, #67 and #68. I have added relevant tests and ensured they pass in Python 3.6 and 3.7. (I've changed the code to allow skipping of the datetime tests.)

thanks!

> From this I have moved all schema code into `mm` so that the marshmallow code is contained. And this is basically just a re-write of `mm`. I've updated `api` to accommodate these changes.
> 
> Otherwise only `core` has been changed. This is because marshmallow now loads schema data into the relevant model. (This can be tested against with code in #58). I'm unsure how you'd want the data to be handled, and so I chose to return early so I don't mangle with `core` too. **please verify this is the intended behavior**.

will address this separately

> This should keep all the previous logic around options, most are in `schema` such as specifying a default, and handling `Optional`. The other special piece of logic is when creating a DC, I kept the `field_many` option. Otherwise `build_type` now builds all the types you'd previously specified, and a couple more, in a recursive form, defaulting to `fields.Field` and a warning.
> 
> I would like to request that we change how you handle the `dataclasses_json.mm_field` so that if the value is `None` we generate one by default. This isn't something you did previously, and so I didn't change this.

yep this makes sense for a future patch

lidatong:
Released in [0.2.2](https://github.com/lidatong/dataclasses-json/releases/tag/v0.2.2)

Peilonrayz:
@lidatong It should be noted that [this line](https://github.com/lidatong/dataclasses-json/blob/master/dataclasses_json/mm.py#L72) should be changed to `return fields.Field(**options)`. As pointed out by @dsvensson otherwise Enum support won't work.

I was waiting for a test case to work against until I committed the fix.

Issue #69, opened 2019-02-27
Recursive encoding of dataclasses does not use overrides


templed:
Hello,

We recently found an issue with the encoding of nested dataclass objects where the nested dataclass has a specific encoder. The problem is that the nested dataclasses's encoder is not used when converting the field value to JSON. A simple example is shown below:

```python
@dataclass_json
@dataclass
class Foo:
    dt: datetime.datetime = field(
        metadata={'dataclasses_json': {
            'encoder': dt.datetime.isoformat,
            'decoder': iso8601.parse_date,
            'mm_field': fields.DateTime(format='iso')
        }})


@dataclass_json
@dataclass
class FooCtr:
    l: List[Foo]


f = Foo.from_json('{"dt": "2019-01-02T12:34:56Z"}')

print(f.to_json())
# {"dt": "2019-01-02T12:34:56+00:00"}

fc = FooCtr([f])

print(fc.to_json())
# {"l": [{"dt": 1546432496.0}]}
```

The timestamp in the `fc` variable is using the default `datetime` encoder in `core.py` which is to make a timestamp. The reason for this is, and a possible fix is given in a pull request.

lidatong:
Addressed by #70. Thanks again

PR #70, opened 2019-02-27
Reorder extraction to allow recursive encoding overrides to work

templed:
Hello,

This PR is meant to resolve issue #69. The problem is that overrides for encoding are only called at the top level, meaning any nested dataclasses do not have their configured overrides applied to them. By moving the `_overrides` and `_override` methods into `_asdict` the recursive calls to each of the fields in the nested dataclasses get their overrides applied.

I'm happy to apply any changes you see fit, and perhaps add some regression tests to ensure this recursive encoding behaviour is recorded.


lidatong:
Thanks for the PR! I'm sorry for the delay

PR #76, opened 2019-03-25
Fix decoding of enums with mixins

### LINKED ISSUE #75: Decoding enums with str mixin fails
Enums can be defined with mixins, for example:

~~~python
class MyStrEnum(str, Enum):
    a = 'aa'
    b = 'bb'
~~~

Decoding values for that enum fails because `_decode_generic` handles them incorrectly:

~~~
AttributeError                            Traceback (most recent call last)

---> 17 ElectionConfig.from_json(jso)

~/.local/share/virtualenvs/ekklesia-portal-GhysHmkk/lib/python3.7/site-packages/dataclasses_json/api.py in from_json(cls, s, encoding, parse_float, parse_int, parse_constant, infer_missing, **kw)
     60                          parse_constant=parse_constant,
     61                          **kw)
---> 62         return _decode_dataclass(cls, kvs, infer_missing)
     63 
     64     @classmethod

~/.local/share/virtualenvs/ekklesia-portal-GhysHmkk/lib/python3.7/site-packages/dataclasses_json/core.py in _decode_dataclass(cls, kvs, infer_missing)
    112             init_kwargs[field.name] = _decode_generic(field.type,
    113                                                       field_value,
--> 114                                                       infer_missing)
    115         elif _issubclass_safe(field.type, datetime):
    116             # FIXME this is a hack to deal with mm already decoding

~/.local/share/virtualenvs/ekklesia-portal-GhysHmkk/lib/python3.7/site-packages/dataclasses_json/core.py in _decode_generic(type_, value, infer_missing)
    149             xs = zip(ks, vs)
    150         else:
--> 151             xs = _decode_items(type_.__args__[0], value, infer_missing)
    152 
    153         # get the constructor if using corresponding generic type in `typing`

/nix/store/6x7zqpvbj07rx837jjvc9cc62qr56fs8-python3-3.7.2-env/lib/python3.7/enum.py in __getattr__(cls, name)
    342         """
    343         if _is_dunder(name):
--> 344             raise AttributeError(name)
    345         try:
    346             return cls._member_map_[name]

AttributeError: __args__
~~~

dpausp:
Fixes #75.

Move up the enum check in _decode_generic to ensure that enums are decoded
properly regardless of any mixin classes.
`MyEnum(str, Enum)` was handled incorrectly as collection before, for
example.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -140,6 +140,9 @@ def _is_supported_generic(type_):
 def _decode_generic(type_, value, infer_missing):
     if value is None:
         res = value
+    elif _issubclass_safe(type_, Enum):
+        # Convert to an Enum using the type as a constructor. Assumes a direct match is found.
+        res = type_(value)
     elif _is_collection(type_):
         if _is_mapping(type_):
             k_type, v_type = type_.__args__
@@ -156,9 +159,6 @@ def _decode_generic(type_, value, infer_missing):
             res = _get_type_cons(type_)(xs)
         except TypeError:
             res = type_(xs)
-    elif _issubclass_safe(type_, Enum):
-        # Convert to an Enum using the type as a constructor. Assumes a direct match is found.
-        res = type_(value)
     else:  # Optional
         type_arg = type_.__args__[0]
         if is_dataclass(type_arg) or is_dataclass(value):
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -13,6 +13,8 @@ class MyEnum(Enum):
     INT1 = 1
     FLOAT1 = 1.23
 
+class MyStrEnum(str, Enum):
+    STR1 = "str1"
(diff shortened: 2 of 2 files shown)


dataclasses_json/core.py
```diff
@@ -140,6 +140,9 @@ def _is_supported_generic(type_):
 def _decode_generic(type_, value, infer_missing):
     if value is None:
         res = value
+    elif _issubclass_safe(type_, Enum):
+        # Convert to an Enum using the type as a constructor. Assumes a direct match is found.
+        res = type_(value)
```
Gulats: @dpausp  should we also consider adding a fallback to `type_[value]` for enum here to be able to parse the enum by name if the parsing by value `type_(value)` fails?

lidatong:
Thank you for the PR! I'm sorry for the delay

PR #84, opened 2019-05-15
Add Jackson style subtype parsing

crutcher:
```
    Declaring a base class which can delegeate from_json()
    to registered subclasses:
    >>> @dataclass_json(parse_as_subtype=True)
    >>> @dataclass(frozen=True)
    >>> class Base:
    >>>     x: int
    >>> 
    >>> @dataclass_json(register_subtype_of=Base)
    >>> @dataclass(frozen=True)
    >>> class Foo(Base):
    >>>     y: str
    >>> 
    >>> b = Base(1)
    >>> foo = Foo(2, 'abc')
    >>>
    >>> b.to_json()  # '{"x": 1}'
    >>> foo.to_json()  # '{"x": 2, "y": "abc", "@type": "Foo"}'
    >>>
    >>> Base.from_json('{"x": 3}') # Base(3)
    >>> Base.from_json('{"x": 2, "y": "abc", "@type": "Foo"}') # Foo(2, "abc")
```

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -4,7 +4,7 @@ from typing import Any, Callable, List, Optional, Tuple, TypeVar, Union
 
 from dataclasses_json.mm import build_schema
 from dataclasses_json.core import (_ExtendedEncoder, _asdict, _decode_dataclass,
-                                   _overrides, _override)
+                                   _overrides, _override, JsonTypeInfo, JsonSubtypeInfo)
 
 A = TypeVar('A')
 B = TypeVar('B')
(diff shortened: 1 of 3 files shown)


crutcher:
Hi; was looking for the subtype mechanism from java's Jackson library, and added it to dataclasses_json.

I'm not sure how to approach this for you; and it doesn't play cleanly with mm (since mm doesn't do this).

george-zubrienko:
I think it should be enough to inherit Base from the DataClassJsonMixin to have this more or less solved automatically?

george-zubrienko:
Closing in favour of #442 - can use for units though

PR #86, opened 2019-05-23
Add decimal.Decimal and typing.NewType support

NeveHanter:
Add decimal.Decimal and typing.NewType support, clean the imports, upgrade marshmallow and bump the version

Signed-off-by: Kamil Bar <nevehanter@gmail.com>

### DIFF
--- a/dataclasses_json/utils.py
+++ b/dataclasses_json/utils.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 from datetime import datetime, timezone
 from typing import Collection, Mapping, Optional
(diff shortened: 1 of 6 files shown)


lidatong:
Thanks so much for the PR and addressing these two issues!

dataclasses_json/utils.py
```diff
@@ -63,11 +64,27 @@ def _isinstance_safe(o, t):
...
+    try:
+        return issubclass(cls, classinfo)
     except Exception:
         return False
-    else:
-        return result
+
+
+def _is_new_type(type_):
+    return inspect.isfunction(type_) and hasattr(type_, "__supertype__")
```
lidatong: this seems like a risky way to determine whether `NewType` as it relies on typing's internal implementation rather than public API, but unless there's a better alternative I'm ok with this as is

PR #87, opened 2019-05-28
Create LICENSE

uriva:


lidatong:
hi, thanks for raising this issue. i'm going to re-license the project under MIT. this should have no material on any users of the library but address some disadvantages of the Unlicense I was not previously aware of

Issue #82, opened 2019-05-07
Union type deserialize to dict from json


denisvlah:
The the Union type on a dataclass cannot be deserialized correctly from json.
Example:

```
from dataclasses import dataclass
from dataclasses_json import dataclass_json
from typing import *


@dataclass_json
@dataclass(unsafe_hash=True)
class Gofy:
    p1: str

@dataclass_json
@dataclass(unsafe_hash=True)
class Foo:
    p2: str

@dataclass_json
@dataclass(unsafe_hash=True)
class Agg:
    inst: Union[Gofy, Foo]


if __name__ == '__main__':
    a = Agg(inst=Foo(p2='p2'))
    json_str = a.to_json()
    a_restored = Agg.from_json(json_str)
    print(type(a_restored.inst) ) # this will print the dict type
    assert a == a_restored # this check fails
    assert isinstance(a_restored.inst, Foo) # this check fails
```

uriva:
+1 did you find any solution?

denisvlah:
My workaround is to deserialize json using dataclass-json lib and after that fix with custom code fields that were deserializwd as dict

lidatong:
hi, thanks to @FaraSeer the latest release (0.2.5) supports union types

fanchunke1991:
> hi, thanks to @FaraSeer the latest release (0.2.5) supports union types

the issues mentioned by @denisvlah actually  exists in version 0.3.6

obi1kenobi:
Just got bitten by this in 0.3.7, and confirmed that the original reproduction test case still points out the bug: 2/2 assertions fail. It appears that #93 may have suffered a regression. Would be happy to help if this is something the maintainers feel would be a good first issue for someone entirely unfamiliar with the codebase of this otherwise awesome library.

FaraSeer:
@fanchunke1991, @obi1kenobi, please use the following way:
```python
Agg.schema().dump(a)
a_restored = Agg.schema().load(json_str)
```
instead of
```python
json_str = a.to_json()
a_restored = Agg.from_json(json_str)
```

obi1kenobi:
Thanks for the suggested workaround. Is there an equivalent workaround for converting from/to dict? Unfortunately, by the time my module gets the data, it has already been decoded from JSON and its response needs to be a dict that will (later on) be encoded as JSON elsewhere.

meandmymind:
[...]

PR #93, opened 2019-06-11
Union type support (#82)

FaraSeer:
- Added new class _UnionField for handling typing.Union (in the serialized data of the classes listed in the Union[...], a new field "__type" will appear)
- Added unit tests

Note:
Only works through the use of the "schema" field:
```
schema = <dataclass>.schema()
schema.dump(<dataclass_instance>)  # for serialization to python dict
schema.dumps(<dataclass_instance>)  # for serialization to json
```
and 
```
schema = <dataclass>.schema()
schema.load(<py_dict>)  # for serialization from python dict
schema.loads(<json_str>)  # for serialization from json
```

`<dataclass>.to_json(<dataclass_instance>)` and `<dataclass>.from_json(<json_str>)` methods are not affected because to_json method does not use marshmallow.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -7,6 +7,7 @@ from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
 from typing import Collection, Mapping, Union, get_type_hints
+from typing_inspect import is_union_type
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


FaraSeer:
@lidatong could you see my pull request?

PR #97, opened 2019-06-20
Added support for from_json() accepting dict in addition to str-like types

USSX-Hares:
### Problem:
Sometimes, you already have loaded json (i.e., from the function result, from json5, from yaml, etc.)
This will help to avoid such construction:
```python
loaded: dict = json5.load(my_opened_file)
data: MyDataclass = MyDataclass.from_json(json.dumps(loaded))
```

Of course, you can always use schema loaders, but there are still redundant calls:
```python
data: MyDataclass = MyDataclass.schema().load(loaded)
```

### Changelog:
 - Added overload for api.from_json() - it now accepts dict in addition to str, bytes and bytearray
 - Added api and annotation tests covering this scenario.
 - Other minor fixes/improvements:
   - `mm.py`: Minor type annotation issue fixed
   - Annotation Tests: added support for mypy of 0.710+
   - `setup.py`: now mypy of 0.710+ is required

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -1,6 +1,6 @@
 import abc
 import json
-from typing import Any, Callable, List, Optional, Tuple, TypeVar, Union, Type
+from typing import Any, Callable, List, Optional, Tuple, TypeVar, Union, Type, overload
 
 from dataclasses_json.mm import build_schema, SchemaType, JsonData
 from dataclasses_json.core import _ExtendedEncoder, _asdict, _decode_dataclass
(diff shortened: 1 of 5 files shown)


lidatong:
Hi, thanks for another PR. i have previous considered making additions to the API to support `dicts`, and i think it makes sense rather than having to go through `.schema()`

rather than overloading `from_json`, i'm going to favor introducing two separate methods `to_dict` and `from_dict`

lidatong:
Addressed by #99. Much thanks again for your help!

PR #100, opened 2019-06-22
Api additions: `to_dict` and `from_dict`

lidatong:
Thanks @USSX-Hares!


USSX-Hares:
@lidatong what about release 0.2.8?

PR #111, opened 2019-07-19
Ensure first section in README.md runs without modifications

r-richmond:
Noticed the first example didn't run when you copy and paste it; this resolves that.


lidatong:
Thanks for catching this!

PR #119, opened 2019-07-31
Union support bugfix

FaraSeer:
Prevents impact on the original serialized data during Union deserialization.
Previously, the field “__type” was being removed from the original data, which made it impossible to deserialize from the original data twice.


FaraSeer:
@lidatong please, could you consider my pull request?

PR #124, opened 2019-08-18
Added test case for self-recursion with Marshmallow Schema

nathan5280:
Added a small example of self-recursion in the README.


lidatong:
Thanks again for this @nathan5280

Issue #146, opened 2019-09-27
Possible issue with timestamp serialization


stevenbedrick:
Apologies if this is a known limitation, but I have run into an issue when using the `.schema().dumps()` functionality. I have encountered an issue with serialization of dataclasses with optional timestamp fields whose value is `None` at the time of serialization. The `.to_json()` method works just fine in this case, but I believe that there is [a bug in line 25 of `dataclasses_json/mm.py`](https://github.com/lidatong/dataclasses-json/blob/03dc2f9bc04c24cfb5d3eded63f6c4176d4779bf/dataclasses_json/mm.py#L25) that is encountering problems when the `value` in question is `None`. 

I have put together a small example script illustrating the issue [here](https://gist.github.com/stevenbedrick/d05970da1a21b60c3cd8fc421d37e90e).

One possible fix would just be to add a check for `None` to the `_TimestampField._serialize()` method; from my read of the [calling code in Marshmallow](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/fields.py#L325), it looks like it's expecting field-specific serializers to be able to handle `None` values in this sort of situation.

FaraSeer:
Hi! If I understood correctly, then the bug is already described here: #108

stevenbedrick:
Yup, that's definitely the same bug. Sorry for the duplicate!

PR #147, opened 2019-09-29
Optional support

### LINKED ISSUE #108: Serializing datetime error
Nice library!

I found one problem when I want serialize datetimes:

```
from datetime import datetime
from typing import Optional

@dataclass
@dataclass_json
class Test:
    id: Optional[int]
    date: Optional[datetime]

test = Test(None, None)
```
If execute:
```
test.to_json()
```
works as expected:
> '{"id": null, "date": null}'

But if  I execute:
```
Test.schema().dump(test)
```
It returns:
> AttributeError: 'NoneType' object has no attribute 'timestamp'

stevenbedrick:
I _think_ this change should fix #108 and #146 by adding a `None` check, and throwing a `ValidationError` in the case of a blank value showing up when it's not allowed according to the schema. As best as I can tell, this is what Marshmallow does in this sort of situation.

### DIFF
--- a/dataclasses_json/mm.py
+++ b/dataclasses_json/mm.py
@@ -25,7 +25,7 @@ from dataclasses_json.utils import (_is_collection, _is_optional,
 class _TimestampField(fields.Field):
     def _serialize(self, value, attr, obj, **kwargs):
         if value is not None:
-            return value.timesimportamp()
+            return value.timestamp()
         else:
             if not self.required:
                 return None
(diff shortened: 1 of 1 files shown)


stevenbedrick:
Oddly, on my local machine, one of the tests is failing (one of the ones about type annotations) and I can't quite figure out why, but the CI above seems happy, so there we go.

dataclasses_json/mm.py
```diff
@@ -23,18 +24,42 @@
...
 class _TimestampField(fields.Field):
     def _serialize(self, value, attr, obj, **kwargs):
-        return value.timestamp()
+        if value is not None:
+            return value.timestamp()
+        else:
+            if not self.required:
+                return None
+            else:
+                raise ValidationError(self.default_error_messages["required"])
```
lidatong: perhaps this code should be factored into a helper?

PR #148, opened 2019-10-05
Use caching to improve performance

konstin:
dataclasses-json is very slow when parsing lists of objects because it introspects the dataclasses every time. This pull request caches the slowest of those calls using python's built-in `lru_cache`, Reducing the decoding time for my test json from 21s to 2.1s. 

There are commits: With 56c3c9d the already failing test keeps failing, while the remainder of the tests passes. With cbe4d87 pytest fails on collecting tests cases due to some type being unhashable.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -6,6 +6,7 @@ from dataclasses import (MISSING, _is_dataclass_instance, fields, is_dataclass)
 from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
+from functools import lru_cache
 from typing import Collection, Mapping, Union, get_type_hints
 from uuid import UUID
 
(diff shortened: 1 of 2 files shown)


lidatong:
Hey there, thanks for raising this and the contribution, I appreicate it a lot. I definitely agree that `dataclasses-json` is doing a rather inefficient thing by traversing the schema on every load.

Coincidentally, I previous considered addressing this with `@lru_cache`, but as you mention there is the hashable input limitation. This can be solved by using a more feature-rich cache implementation (e.g. `cachetools`). Specifically, `cachetools` supports a `key` function that allows you to specify how to derive a hashable key from your object.

Please feel free to implement, or I'm also happy to take this over the line. Thanks again!

konstin:
Unfortunately I don't have the time to look into `cachetools`.

lidatong:
Superseded by #175

PR #150, opened 2019-10-11
Enable mypy to find dataclasses_json when installed

gward:
With this change, mypy can find dataclasses_json and can make use of its type annotations.


lidatong:
Thanks for the contribution!

PR #156, opened 2019-11-25
Fix `from_dict` behaviour when using Dict annotations

eqperes:
`from_dict` used to break for dataclasses using `Dict[Any, Any]` or `dict` as typing annotations. 

That was because `Any` cannot be instantiated and because there is no `__args__ in `dict`.


milesgranger:
It appears this will close #162 :+1:

eqperes:
@lidatong Sorry for my late answer. I have added some unittests for fields with `dict` and `Dict`.

eqperes:
Thank you @lidatong  👍

PR #164, opened 2020-01-06
Support usage of a custom encoder on nested dataclasses

whisust:
Currently, if we use a custom encoder on a nested dataclass, it is not called during serialization.

This pull request fixes it, with associated test cases.

As my use case (and thus the unit test 😄) was to strip `None` values at serialization, I was wondering if this was a common enough use case to add a boolean flag in the dataclass_json configuration to avoid writing a custom Encoder:

```python
def dataclass_json(_cls=None, *, letter_case=None,
                   undefined: Optional[Union[str, Undefined]] = None,
                   skip_none: Optional[bool] = None):
```

If that's the case, I shall add this case to the `config` and see from there how to use the `cls.dataclass_json_config` at serialization time

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -288,7 +288,10 @@ def _asdict(obj, encode_json=False):
     if _is_dataclass_instance(obj):
         result = []
         for field in fields(obj):
-            value = _asdict(getattr(obj, field.name), encode_json=encode_json)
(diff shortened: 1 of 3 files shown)


matt035343:
Hi @whisust, thanks for the contribution. Please update to match newest version on master and resolve conflicts, then we can proceed with a review

PR #168, opened 2020-01-20
Customized JSON econder/decoder function

phakiller:
Modifies 'DataClassJsonMixin' class methods to accept a customized JSON econder/decoder function.
Would be great this modification, in my case I can use `orjson` lib instead default Python default lib.


phakiller:
@lidatong Can you ratify this PR, please?

lidatong:
I just realized the `_ExtendedEncoder` is coupled with the standard library `json` module. Until we separate the two, I don't configuring the `json_module` will work properly. I'll create a separate issue

PR #178, opened 2020-02-20
datetime.timedelta support

aidarvaleev1998:
Added support of `datetime.timedelta`:
encode: `timedelta.total_seconds`, decode: `lambda s: timedelta(seconds=s)`

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -6,7 +6,7 @@ from collections import namedtuple
 from dataclasses import (MISSING, _is_dataclass_instance, fields,
                          is_dataclass  # type: ignore
                          )
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from decimal import Decimal
 from enum import Enum
 from typing import Any, Collection, Dict, Mapping, Union, get_type_hints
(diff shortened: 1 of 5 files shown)


lidatong:
Hi, I'm not sure `timedelta` is a common enough field to warrant support in the library itself. I do want to avoid supporting a proliferation of types. My original intent was to support only the JSON types that the official Python `json` module supported. I then added a few additional types that were extremely common, e.g. UUID or datetime. The goal is avoiding a "slippery slope" of an if-ladder of many supported types and keeping the library small.

I'm not sure exactly how to address the fact there are these "in-between" types, that are shared library types but not exactly. Maybe a "dataclasses-json-extensions" module?

That said, feel free to respond with your thoughts on this case in particular and thank you for the contribution, I do appreciate it a lot.

aidarvaleev1998:
Hello, the problem was that I had a dataclass inherited from another dataclass which has optional fields, and previously this child dataclass had only optional fields, but to extend your library dataclasses.field(metadata) assignment is required which is forbidden in pyrhon due to the presence of optional fields in the parent. Is it possible now to have optional fields of extended datatypes or I still need dataclasses.field(metadata) assignment?

aidarvaleev1998:
Took a look at https://github.com/lidatong/dataclasses-json/pull/189 and understood that the answer for the above question is yes.

lidatong:
Sorry I didn't get back to you. Glad you figured it out!

PR #184, opened 2020-02-25
fix cache using id -> hash conflicts for classes with the same name

Germandrummer92:
Unfortunately my previous cache pr didn't work when it came to overwriting classes with the same name, see the test case I added that fails on the previous version and passes now.

This is because the previous memory address is free again after overwriting a class with a class with the same name and a new assignment will (possibly ? not sure about cpythons internals here 😄 ) reuse it.

This uses repr and an injected class field instead.

Performance on the load_test (I committed that feel free to remove it/move it around):

No cache:
```
/Applications/miniconda3/envs/dataclasses-json/bin/python /Users/ddraper/dev/dataclasses-json/load_test.py
It took 42.135189056396484s to convert to dict
It took 32.14417791366577s to convert from dict
```

With cache and _dataclass_hash as the key:
```
/Applications/miniconda3/envs/dataclasses-json/bin/python /Users/ddraper/dev/dataclasses-json/scripts/load_test.py
It took 40.158539056777954s to convert to dict
It took 19.082113027572632s to convert from dict

Process finished with exit code 0
```

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -57,7 +57,14 @@ confs = ['encoder', 'decoder', 'mm_field', 'letter_case']
 FieldOverride = namedtuple('FieldOverride', confs)
 
 
-@cached(cache=_user_overrides_cache, key=repr)
+def _hash_key(cls):
+    if hasattr(cls, '_dataclass_hash'):
+        return cls._dataclass_hash
+    else:
+        return repr(cls)
+
+
+@cached(cache=_user_overrides_cache, key=_hash_key)
 def _user_overrides_or_exts(cls) -> Dict[str, FieldOverride]:
     overrides = {}
     # overrides at the class-level
@@ -128,7 +135,7 @@ def _decode_letter_case_overrides(field_names, overrides):
     return names
 
 
-@cached(cache=_get_type_hints_cache, key=repr)
+@cached(cache=_get_type_hints_cache, key=_hash_key)
 def _get_type_hints_cached(cls):
     return get_type_hints(cls)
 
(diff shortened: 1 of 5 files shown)


Germandrummer92:
@lidatong @RunOrVeith

dataclasses_json/core.py
```diff
@@ -57,7 +57,7 @@ def default(self, o) -> Json:
 FieldOverride = namedtuple('FieldOverride', confs)
 
 
-@cached(cache=_user_overrides_cache, key=id)
+@cached(cache=_user_overrides_cache, key=repr)
```
RunOrVeith: I am not quite sure why repr fixes the problem.
The default `__repr__` implementation returns basically `f"<{__name__}.{self.__class__.__name__} object at {id(self)}>`, so if the class is overwritten with the same name and the variable naming stuff is done as before, this will result in the same repr as well, or am I mistaken?

```
class Test():
...     pass
... 
Test().__repr__()
'<__main__.Test object at 0x7f458df2ab50>'
# Overwrite it
class Test():
...     pass
... 
Test().__repr__()
'<__main__.Test object at 0x7f458df2ab50>'
```

RunOrVeith: Same for the other caches

Germandrummer92: yeah true, I will try to find another solution

Germandrummer92: fixed with an injected _dataclass_hash field. this now works correctly for all cases I can think of...

lidatong:
Hey, thanks for following up on this with a fix. I've reverted the caching optimization for now because of some bugs users have been reporting.

While I think this PR could address some of the current issues we're experiencing with caching, I'm not so sure it accounts for everything. There's a lot of potential test cases to implement to gain confidence it works (and the test cases will have to be sensitive to the timing, ordering and execution path of the code). The trouble with caching is all the complexities it introduces: hashing / identity, longevity, invalidation, memory usage, etc.

I've been thinking about how we might improve this lib's performance without caching, and one alternative is code generation. The code for schema is generated once on startup. It's "compiled", so to speak.

Code generation is the approach `dataclasses` itself, and schema compilation is generally the way I've observed codec libraries implemented. Please feel free to respond with your thoughts.

Issue #181, opened 2020-02-23
Add documentation for how to use global_config


lidatong:
`test_global_config.py` has a usage example

george-zubrienko:
Can be closed now?

PR #189, opened 2020-03-11
Documentation: added example of global_config

nr23:
The "Extended" section in the documentation has the wrong example. This fixes the example so that it uses `global_config` as mentioned in #181


george-zubrienko:
@nr23 could you please sync branch so we can merge it after CI runs :)

wvdemeer:
done

PR #196, opened 2020-03-20
handle field_type is datetime and field_value is str

nurettin:
convert string to datetime if field_type is datetime and field_value is string.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -12,9 +12,8 @@ from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
 from typing import Any, Collection, Mapping, Union, get_type_hints
-from uuid import UUID
-
 from typing_inspect import is_union_type  # type: ignore
+from uuid import UUID
 
 from dataclasses_json import cfg
 from dataclasses_json.utils import (_get_type_cons,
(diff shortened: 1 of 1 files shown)


kevinmu:
I am in need of this functionality :) Would like to get your thoughts on this @lidatong 

For this diff, what do you think about using `dateutils.parse` instead of `strptime`? In python 3.7 there is also `datetime.fromisoformat`

lidatong:
Hi, please, take a look at [this section in the docs](https://github.com/lidatong/dataclasses-json#overriding) on overriding the `decoder` for a certain field. Part of the goal is for the library to minimize the list of supported types to avoid code bloat. This also adds the overhead of checking for every `str` field whether it happens to be an ISO-formatted value -- which doesn't make sense

nurettin:
> Hi, please, take a look at [this section in the docs](https://github.com/lidatong/dataclasses-json#overriding) on overriding the `decoder` for a certain field. Part of the goal is for the library to minimize the list of supported types to avoid code bloat. This also adds the overhead of checking for every `str` field whether it happens to be an ISO-formatted value -- which doesn't make sense

Hi, I agree that keeping supported types to a minimum is good, but this only adds checking if `field_type` is datetime, not every time a string is encountered. So it does make a lot of sense.

PR #197, opened 2020-03-23
Correct Example for Different name Encode/Decode

r-richmond:
Remove incorrect/old code in example


lidatong:
Thanks for fixing!

Issue #206, opened 2020-04-08
Pretty errors when deserializing objects


eblis:
It would be nice if a "pretty" exception was raised when a particular json cannot be mapped to a @dataclass-json object, instead of raising `KeyError` exceptions.

Right now if the json string cannot be mapped a KeyError exception is raised, saying that a particular field isn't available, but this field might be from a child of a child of the root object and it might be hard to determine which field it's taking about just by the name

lidatong:
yep, this would be useful -- i think this is a good opportunity to take a stab at an initial error hierarchy (deserialization, serialization, etc.), as `KeyError` is just one potential overly-general error, setting the base to a feature branch for now rather than `master`

Mtihc:
I'm not sure how this library works internally. But I'd just like to add that, not only `from_json`, but also `schema().loads` raises a `KeyError`.

```python
from typing import Optional
from dataclasses import dataclass
from dataclasses_json import DataClassJsonMixin

@dataclass
class MyModel(DataClassJsonMixin):
    required_value: str
    optional_value: Optional[str] = None

MyModel.from_json('{}') # KeyError
MyModel.schema().loads('{}') # KeyError

```

I would expect that `MyModel.schema().loads('{}')` raises:
```
marshmallow.exceptions.ValidationError: {'required_value': ['Missing data for required field.']}
```

I wanted to mention this because I see you speaking of error hierarchies. My first thought was: "what hierarchy? I just want the `marshmallow.exceptions.ValidationError`!"

tuukkamustonen:
> I would expect that `MyModel.schema().loads('{}')` raises:
> 
> ```
> marshmallow.exceptions.ValidationError: {'required_value': ['Missing data for required field.']}
> ```

This is really something you'd expect the library to do out of box.

george-zubrienko:
Target of v1 API #442 - take a look.

PR #207, opened 2020-04-09
Added custom exception (inheriting from KeyError) when deserialization fails

eblis:
I have added a quickly put together patch that helps with issue https://github.com/lidatong/dataclasses-json/issues/206, it will raise a custom exception (that inherits from KeyError from clients that were catching that exception) and reports the actual object that failed and the field it failed for.


eblis:
Hmm, why did the second try gent cancelled, I think it should have passed the linter this time

dataclasses_json/errors.py
```diff
@@ -0,0 +1,3 @@
+class DataclassSerializationException(KeyError):
```
lidatong, line 1: perhaps `class DataclassJsonError(Exception)` instead, to make it clear it's coming from this package, not dataclasses itself -- user-defined exceptions [typically inherit from `Exception`](https://docs.python.org/3/tutorial/errors.html#tut-userexceptions)

lidatong:
see _Originally posted by @lidatong in https://github.com/lidatong/dataclasses-json/issues/206#issuecomment-617468714_

dataclasses_json/errors.py
eblis: It's weird not inheriting from `Exception` but I wanted to inherit from `KeyError` so that any client code that might have caught the old exception would still catch the new one.
I feel like inheriting from `Exception` would be a bit of a breaking change .. fine by me, but not sure if everyone would feel that way.

lidatong: gotcha, that's a good thought and thanks for clarifying

Issue #213, opened 2020-04-30
Bug in enum test cases: schema can't handle int


RunOrVeith:
The tests inside `test_enum.TestValidator` have relied on a wrong operator binding order and thus
two cases that are actually not valid marshmallow were tested to be valid.

The issue is the following line:
```
assert not res == is_valid
```
`res` here is the output of `schema.validate()`, which returns a dict of validation errors.

However, `not` as lower precedence than `==` ([see here](https://www.programiz.com/python-programming/precedence-associativity)), so the statement was treated as
```
assert not (res == is_valid)
``` 
which was `True`no matter what, since we compared a dict to a bool.

After fixing the precedence, two test cases fail: `TestValidator.test_data_with_enum` `(1, True)` and `(1.23, True)`, because marshmallow does in fact not allow int enum values (there is a marshmallow_enum in the ecosystem [here](https://github.com/justanr/marshmallow_enum)).

I will enclose a pull request that tests that these cases are actually `False`.
I found this while implementing a feature that allows enums as dict keys, which I will also link.
@lidatong

PR #215, opened 2020-04-30
Allow enum dict keys

RunOrVeith:
This pull requests allows enums as dict keys.
Since marshmallow does not support anything but string enums by default (see #213 ), 
the marshmallow code for enum dict keys also only works with strings.
I've implemented it so that it works for other stuff (e.g. `float, int`) when using `from_json`, `to_json`,
if that is not desired let me know.

This means you can now do:

```
@dataclass_json
@dataclass(frozen=True)
class DataWithEnumKeys:
    name: str
    enum_dict: Dict[MyEnum, str]


d_keys = DataWithEnumKeys(name="name1", enum_dict={MyEnum.STR1: "str_test",
                                                   MyEnum.INT1: "int_test"})
d_keys.to_json() # '{"name": "name1", "enum_dict": {"str1": "str_test", "1": "int_test"}}'
DataWithEnumKeys.from_json(d_keys.to_json()) == d_keys  # True

```

@lidatong

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -1,19 +1,22 @@
 import copy
 import json
 import warnings
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 # noinspection PyProtectedMember
-from dataclasses import (MISSING, _is_dataclass_instance, fields,
+from dataclasses import (MISSING,
+                         _is_dataclass_instance,
+                         fields,
                          is_dataclass  # type: ignore
                          )
 from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
-from typing import Any, Collection, Dict, Mapping, Union, get_type_hints
+from typing import Any, Collection, Mapping, Union, get_type_hints
 from uuid import UUID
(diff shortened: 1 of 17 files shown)


RunOrVeith:
@lidatong

dataclasses_json/core.py
```diff
@@ -234,9 +234,18 @@ def _decode_generic(type_, value, infer_missing):
     if value is None:
         res = value
     elif _issubclass_safe(type_, Enum):
-        # Convert to an Enum using the type as a constructor.
-        # Assumes a direct match is found.
-        res = type_(value)
+        # we got the enum value
+        for enum_member in type_:
+            # We rely on the user that the enum values are unique
```
lidatong, line 239: I don't believe this is necessarily true -- enum names are not necessarily one-to-one with values. There is a decorator `@unique` that enforces it, but the standard library does allow for it

ole-tidal:
I would like to have this feature as well. Super handy to have especially when coming from Jackson in java.

PR #216, opened 2020-05-15
Fix deprecation warning

### LINKED ISSUE #185: DeprecationWarning for 'encoding' argument with json.loads
I get the following deprecation warning when using from_json() method in **python 3.8**:

```
/usr/local/lib/python3.8/site-packages/dataclasses_json/api.py:72: DeprecationWarning: 'encoding' is ignored and deprecated. It will be removed in Python 3.9
  kvs = json.loads(s,
```

Happy to submit a PR to remove this field if you want :)

stevenj:
Fixes the deprecation warning for 3.9 "dataclasses_json/api.py:72: DeprecationWarning: 'encoding' is ignored and deprecated. It will be removed in Python 3.9"

Fixes #185 

This field was removed from json.loads in 3.1


stevenj:
I Don't know what the story is with failing flake8 tests, nothing was changed that could effect that and running flake8 does not show any errors in the file i touched, but there are a few in unrelated files.

RunOrVeith:
Not sure if we want a breaking change without a warning.
We could keep the `encoding` field in `from_json`, and issue a warning that it is being ignored if it is not `None`. 
I personally would also be ok with just kicking it out, since the deprecation warning has been there for a while.
But I am not sure, what do you think @lidatong.

Regarding the flake8 error, I see two options:
- Replace `((lambda x: x) if key_type is type(None) or key_type == Any` with `((lambda x: x) if isinstance(key_type, type(None)) or key_type == Any`
- Ignore the error by adding # pylint: disable=E721 since it is wrong here (the error is there to prevent people from checking against a class type and forgetting about subclasses, which we are sure not to do here because we just check against NoneType).

I'd prefer the first version. I am not sure why it fails now, but it'd be great if you could incorporate this small fix :)

lidatong:
Hi, thanks for making the fix to the warning.

@RunOrVeith it's a good point on the backwards compatibility, but `from_json` has a catch-all for additional keyword arguments with `**kw` so it will remain compatible with passes `encoding`

lidatong:
Reason for the flake8 error
> E721 do not compare types, use 'isinstance()'

If you click `Details` you can see the reasons for the failing builds

stevenj:
Is this going to get merged or are you waiting on something from me?

lidatong:
I'll address the build error and push out a release tomorrow. Thanks for following up.

PR #221, opened 2020-05-23
Allow fields to be excluded from encoding

### LINKED ISSUE #200: Omitting fields when serializing
Is it possible to omit fields when `to_dict` or `to_json` is invoked?  My use-case is I have fields with sensitive information that I only want available to the internal parts of my application, and would like to ensure they're never pushed to my end-users.

Right now, I'm accomplishing it by overriding `to_dict` as follows:

```python
@dataclass
class MyObject(DataClassJsonMixin):
    """An object."""

    id_: uuid.UUID = field(metadata=config(field_name="id"))
    name: str
    secret: str
    another_secret: str

    def to_dict(  # type: ignore[override]
        self, omit_sensitive=True, **kwargs
    ) -> Dict[str, "dataclasses_json.core.Json"]:
        """Serialize the dataclass, with an optional ability to omit sensitive/internal fields."""
        serialized = super().to_dict(**kwargs)
        if omit_sensitive:
            for field_name in list(serialized.keys()):
                if field_name in ("secret", "another_secret"):
                    del serialized[field_name]
        return serialized
```

Ideally, this would be something I could define as part of the `field()`.  Is this currently possible with dataclass-json, and, if not, would this be something that you would consider for future functionality?

arusahni:
Introduce an `encode_action` field config parameter. When set to exclude
a field, the encoder will not include it in the encoded dictionary.

The config parameter can also accept a callback that receives the
key-value pair being encoded, allowing for more-granular control of the
exclusion check.

I'm very much open to modifying this to fit the goals of the library.  Also, I'm not sure how one goes about updating the documentation, but, if you point me in the right direction, I'm more than happy to update it as well.

Fixes #200

### DIFF
--- a/dataclasses_json/cfg.py
+++ b/dataclasses_json/cfg.py
@@ -1,9 +1,23 @@
 import functools
-from typing import (Callable, Dict, Optional, Union)
+from typing import Callable, Dict, NamedTuple, Optional, TypeVar, Union
 
 from marshmallow.fields import Field as MarshmallowField
 
 from dataclasses_json.undefined import Undefined, UndefinedParameterError
+
+T = TypeVar("T")
+
+
+class _Exclude(NamedTuple):
+    """
+    Whether or not the field should be excluded when encoded
+    """
+
+    ALWAYS: Callable[[T], bool] = lambda _: True
+    NEVER: Callable[[T], bool] = lambda _: False
+
+
+Exclude = _Exclude()
 
 
 # TODO: add warnings?
@@ -35,7 +49,9 @@ def config(metadata: dict = None, *,
            mm_field: MarshmallowField = None,
            letter_case: Callable[[str], str] = None,
            undefined: Optional[Union[str, Undefined]] = None,
-           field_name: str = None) -> Dict[str, dict]:
+           field_name: str = None,
+           exclude: Optional[Callable[[str, T], bool]] = None,
+           ) -> Dict[str, dict]:
     if metadata is None:
         metadata = {}
 
@@ -74,5 +90,8 @@ def config(metadata: dict = None, *,
             undefined = Undefined[undefined.upper()]
 
         lib_metadata['undefined'] = undefined
+
+    if exclude is not None:
+        lib_metadata['exclude'] = exclude
 
     return metadata
--- a/dataclasses_json/__init__.py
+++ b/dataclasses_json/__init__.py
@@ -2,5 +2,5 @@
 from dataclasses_json.api import (DataClassJsonMixin,
                                   LetterCase,
                                   dataclass_json)
-from dataclasses_json.cfg import config, global_config
+from dataclasses_json.cfg import config, global_config, Exclude
 from dataclasses_json.undefined import CatchAll, Undefined
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -48,7 +48,7 @@ class _ExtendedEncoder(json.JSONEncoder):
 
 
 def _user_overrides_or_exts(cls):
-    confs = ['encoder', 'decoder', 'mm_field', 'letter_case']
+    confs = ['encoder', 'decoder', 'mm_field', 'letter_case', 'exclude']
     FieldOverride = namedtuple('FieldOverride', confs)
 
     global_metadata = defaultdict(dict)
@@ -94,6 +94,11 @@ def _encode_overrides(kvs, overrides, encode_json=False):
     override_kvs = {}
     for k, v in kvs.items():
         if k in overrides:
+            exclude = overrides[k].exclude
+            # If the exclude predicate returns true, the key should be
+            #  excluded from encoding, so skip the rest of the loop
+            if exclude and exclude(v):
+                continue
             letter_case = overrides[k].letter_case
             original_key = k
             k = letter_case(k) if letter_case is not None else k
--- a/tests/test_exclude.py
+++ b/tests/test_exclude.py
@@ -1,0 +1,51 @@
+from dataclasses import dataclass, field
+
+from dataclasses_json.api import DataClassJsonMixin, config
+from dataclasses_json.cfg import Exclude
+
+
+@dataclass
+class EncodeExclude(DataClassJsonMixin):
+    public_field: str
+    private_field: str = field(metadata=config(exclude=Exclude.ALWAYS))
+
+
+@dataclass
+class EncodeInclude(DataClassJsonMixin):
+    public_field: str
+    private_field: str = field(metadata=config(exclude=Exclude.NEVER))
+
+
+@dataclass
+class EncodeCustom(DataClassJsonMixin):
+    public_field: str
+    sensitive_field: str = field(
+        metadata=config(exclude=lambda v: v.startswith("secret"))
+    )
+
+
+def test_exclude():
+    dclass = EncodeExclude(public_field="public", private_field="private")
+    encoded = dclass.to_dict()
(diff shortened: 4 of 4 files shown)


dataclasses_json/cfg.py
```diff
@@ -1,11 +1,20 @@
...
-from typing import (Callable, Dict, Optional, Union)
+from enum import auto, Enum
+from typing import (Any, Callable, Dict, Optional, Union)
 
 from marshmallow.fields import Field as MarshmallowField
 
 from dataclasses_json.undefined import Undefined, UndefinedParameterError
 
 
+class EncodeAction(Enum):
```
lidatong, line 10: rather than taking an `enum` let's simplify the API and just have it always expect a predicate

tests/test_encode_action.py
```diff
@@ -0,0 +1,43 @@
+from dataclasses import dataclass, field
```
lidatong, line 1: good tests

arusahni:
The reason I avoided a predicate for the `NEVER` state was an optimization (perhaps premature) to avoid the overhead of a function call for the default path. I do like the symmetry of complementary predicates, however, so I'm fine switching to that.

To ensure I understand what you're suggesting, should I drop the enum altogether and accept a lambda predicate as the config parameter value? Or, should I replace the `auto()` enum values with the predicates?

arusahni:
After sleeping on it, I realized that my concerns weren't well-founded as the predicate is only invoked if explicitly declared.  I've pushed an update that:

* Renames the configuration parameter to `exclude`
* Exposes two helper predicates: `ALWAYS` and `NEVER`

lidatong:
> To ensure I understand what you're suggesting, should I drop the enum altogether and accept a lambda predicate as the config parameter value?
Yep, though the library can still expose a constant to cover the common-case of `lambda _: True`

> should I replace the enum implementation with a namedtuple with the predicates?
Hm not following this -- why a `namedtuple`?

Thanks again

dataclasses_json/cfg.py
```diff
@@ -1,10 +1,23 @@
...
+from typing import (Any, Callable, Dict, NamedTuple, Optional, Union)
 
 from marshmallow.fields import Field as MarshmallowField
 
 from dataclasses_json.undefined import Undefined, UndefinedParameterError
 
+ExcludePredicate = Callable[[str, Any], bool]
+
+
+class _Exclude(NamedTuple):
```
lidatong, line 11: also commented [here](https://github.com/lidatong/dataclasses-json/pull/221#issuecomment-633634103) but why a `NamedTuple`?

dataclasses_json/cfg.py
```diff
@@ -1,10 +1,23 @@
 import functools
-from typing import (Callable, Dict, Optional, Union)
+from typing import (Any, Callable, Dict, NamedTuple, Optional, Union)
 
 from marshmallow.fields import Field as MarshmallowField
 
 from dataclasses_json.undefined import Undefined, UndefinedParameterError
 
+ExcludePredicate = Callable[[str, Any], bool]
```
lidatong, line 8: i'd favor making a unary predicate that takes in the field value, rather than the both field name and value. my reasoning is the field name is constant across instances vs. the value which varies

arusahni: By supplying both the key and value, the library could acquire an interesting layer of runtime indirection, allowing for externally-defined behavior without altering the dataclass signature.

That being said, if it's not something desired, I'm totally fine updating this to use a unary predicate.

dataclasses_json/cfg.py
arusahni: With an enum, the `NEVER` and `ALWAYS` members are of the `Exclude` type. With a namedtuple, they are of the `ExcludePredicate` type. This keeps the type signature of `config(exclude=)` simple, accepting `Optional[ExcludePredicate]` instead of `Optional[Union[Exclude,ExcludePredicate]]`.

lidatong: Got it. What about solving the problem with this pattern?

```python
class Exclude:  #avoids `Enum` or `NamedTuple` altogether
      ALWAYS = lambda _: True
```

I'd favor keeping it as a `Callable` type, rather than introducing a type alias (mainly to avoid adding to the library's API)

Have you considered having the `Callable` signature accept a `TypeVar` instead of `Any`?

This way the type argument is captured and type-checked by users optionally using `mypy` or similar tooling

dataclasses_json/cfg.py
lidatong: > avoids `Enum` or `NamedTuple` altogether
 But here I'd favor to keep it simple with a plain `Callable[[T], bool]`

dataclasses_json/cfg.py
arusahni: * NamedTuples are immutable, while classes aren't. I don't have strong opinions here, so let me know which sits better and I'll make the change.
* Agreed regarding keeping the API clean, I'll remove the alias.
* I think mypy is capable of narrowing `Any` based on what it sees at analysis-time, with a TypeVar only being useful to enable overloading, but using the TypeVar won't hurt things (and looks nicer), so I'm fine with that. I'll make the change and push it up :+1:

dataclasses_json/cfg.py
arusahni: Not quite metaprogramming, but definitely hosting configuration outside of the dataclass itself so one can consolidate logic around `to_dict()` `to_json()` without having to subclass or add additional layers of handling.

An example - if I was returning the dataclass from a web endpoint, and wanted a user to be able to supply which fields they wanted as parameters, I could do something like:

```python
from flask import request

def is_report_field(k, v):
    return k in request.parameters.split(",")

# ...

@dataclass
class Report(DataClassJsonMixin)
    summary: str = field(metadata=config(exclude=is_report_field))
```
While it isn't a use-case I care about, if we were to go with the value-only predicate initially, and then later decide that the key name would be nice to have, this would be difficult to manage without a breaking API change. 

I think that articulates my thinking. Let me know if this is something we don't wish to support - if not, I'll push up a unary-predicate version.

lidatong: Thanks for the example, I think I understand now. I'm not quite compelled it's enough of a use-case to warrant support in the lib, and there's some overhead to users having to write `lambda _, v: ...` vs. just passing in a predicate

arusahni:
Great! I'll push the predicate changes out this evening.

dataclasses_json/cfg.py
arusahni: > It's just a static constant that scoped within a class (stylistically), could be module-level

We do get some niceties out of the box - attributes cannot be overwritten (raises an AttributeError), and new attributes cannot be added at runtime.  With Python 3.8, we do get closer with the `Final` type annotation, but that's not enforced at runtime.

arusahni:
I've pushed the unary predicate update. Thanks for the review, and for accepting this into the library! It's been immensely useful.

lidatong:
Great, thank you for implementing this

dataclasses_json/cfg.py
lidatong: Understand your point, but I think a namedtuple in which only a single instance is ever intended to exist adds confusion and obscures -- I'd rather achieve this functionality of "runtime enforcement" by overriding the dunder methods. But I'm going to make a call to simplify here and just have an uppercase constant that is constant by convention

jcmuddle:
Thanks for making this PR. I’m excited to try it out!

PR #225, opened 2020-06-01
Move FieldOverrides creation outside of _user_overrides_or_exts

justinaustin:
Hello! I have a fairly large dataset (~7 million entries) that I am deserializing with this library, but unfortunately deserialization has become a bottleneck. After a bit of profiling, it looks like the majority of runtime is spent in `_user_overrides_or_exts` within `core.py`. More specifically, `_user_overrides_or_exts` creates a `FieldOverride` `NamedTuple` every time the method is called.

While the namedtuple creation may seem trivial, the runtime adds up when deserializing a large dataset. Below is a sample from my `cProfile` run:
```    Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  7171820    8.270    0.000 5588.569    0.001 schema.py:691(load)
  7171820   45.781    0.000 5580.299    0.001 schema.py:807(_do_load)
  7171820   15.652    0.000 5046.081    0.001 schema.py:1071(_invoke_load_processors)
 14343640   27.048    0.000 5030.430    0.000 schema.py:1192(_invoke_processors)
  7171820   12.661    0.000 4999.333    0.001 mm.py:325(make_instance)
  7171820  154.023    0.000 4986.672    0.001 core.py:101(_decode_dataclass)
  7171821   89.148    0.000 2775.553    0.000 core.py:47(_user_overrides)
  7172281  114.348    0.000 2603.727    0.000 __init__.py:357(namedtuple)
```

Is there a reason I'm missing that `FieldOverrides` is created within the `_user_overrides_or_exts` function instead of once at the top of the file? By only creating `FieldOverrides` once, this saves a ton of runtime and improves the performance of my current use case by more than an order of magnitude.

Thank you!

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -11,7 +11,6 @@ from dataclasses import (MISSING,
 from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
-from functools import lru_cache
 from typing import Any, Collection, Mapping, Union, get_type_hints
 from uuid import UUID
 
@@ -25,6 +24,9 @@ from dataclasses_json.utils import (_get_type_cons,
                                     _issubclass_safe)
 
 Json = Union[dict, list, str, int, float, bool, None]
+
+confs = ['encoder', 'decoder', 'mm_field', 'letter_case', 'exclude']
+FieldOverride = namedtuple('FieldOverride', confs)
 
 
 class _ExtendedEncoder(json.JSONEncoder):
@@ -48,11 +50,7 @@ class _ExtendedEncoder(json.JSONEncoder):
         return result
 
 
-@lru_cache(maxsize=128)
 def _user_overrides_or_exts(cls):
-    confs = ['encoder', 'decoder', 'mm_field', 'letter_case', 'exclude']
-    FieldOverride = namedtuple('FieldOverride', confs)
-
     global_metadata = defaultdict(dict)
     encoders = cfg.global_config.encoders
     decoders = cfg.global_config.decoders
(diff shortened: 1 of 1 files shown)


aronszanto:
A huge +1 to this, it makes some serialization routines an order of magnitude faster on my machine. Also, judicious use of a library like https://pypi.org/project/memoization/ could cache the result of the whole function (and maybe others throughout this library) for further speedups.

justinaustin:
If caching via the `memoization` library is something that you are open to, I'm happy to update this PR (or submit another one) adding it to various functions within this package!

lidatong:
Hi, thanks for finding this optimization and contributing the PR!

There was a previous effort to add caching (see #175). It was introduced in 0.4.1, but I had to revert it in 0.4.2 when a lot of users reported strange bugs.

Performance in general is on my radar as things to tackle next, as this library gains traction, and the top of a 1.0 release checklist.

In general after some thought I don't think caching / memoization is the right way to tackle this. A few reasons why:
1. it requires careful thought about how it behaves under concurrency, specifically with respect to memory visibility
2. could have a big memory footprint on large codebases with a lot of composite dataclasses, and potentially duplicated across threads!
3. immutability -- should the cached object be mutable / how can we protect it from changes?

Instead, I think an approach involving code generation is the way to go -- similar to how the `dataclasses` core module itself is implemented. When you think about it, a schema is only generated once and known at "module-load time". In other languages we might call this "compile-time". We can see the code-generation approach utilized in codec schema libraries in other languages, be it `json` or even other data-interchange formats like  `protobuf`

Going this route, the schema now is loaded as just more code, so to speak, instead of living in memory

aronszanto:
Awesome! Thanks @lidatong and @justinaustin :) Super excited to use this as soon as it's released.

PR #231, opened 2020-06-10
Fix type annotation for the exclude config arg

arusahni:
It seems that, when we dropped the second predicate parameter, we missed this occurrence.

### DIFF
--- a/dataclasses_json/cfg.py
+++ b/dataclasses_json/cfg.py
@@ -50,7 +50,7 @@ def config(metadata: dict = None, *,
            letter_case: Callable[[str], str] = None,
            undefined: Optional[Union[str, Undefined]] = None,
            field_name: str = None,
-           exclude: Optional[Callable[[str, T], bool]] = None,
+           exclude: Optional[Callable[[T], bool]] = None,
            ) -> Dict[str, dict]:
     if metadata is None:
         metadata = {}


dataclasses_json/cfg.py
```diff
@@ -50,7 +50,7 @@ def config(metadata: dict = None, *,
            letter_case: Callable[[str], str] = None,
            undefined: Optional[Union[str, Undefined]] = None,
            field_name: str = None,
-           exclude: Optional[Callable[[str, T], bool]] = None,
+           exclude: Optional[Callable[[T], bool]] = None,
```
jackgene, line 53: Should the function input be a `core.Json`, instead of an arbitrary `T`?

```suggestion
           exclude: Optional[Callable[[Json], bool]] = None,
```

jackgene: `from core import Json`

jackgene (commented):
Typing question.

dataclasses_json/cfg.py
andersk: We do convert `tuple` and `set` to `list`, and nested dataclasses to `dict`, before passing them to `exclude`. However, we pass `datetime`, `UUID`, and `Decimal` objects directly to `exclude` as-is. (Is that discrepancy intended?)

PR #241, opened 2020-08-03
Move exclude logic to improve performance

stafak:
Hi!

I'm dealing with dataclasses that contain lists and numpy arrays that are very large but are excluded when encoding to JSON. Unfortunately, these lists and arrays make the serialization take a very long time despite being excluded. This appears to be due to _asdict running recursively on each list item before the `exclude` property is checked.

Unless I'm missing something, it's not possible to control this from outside the library, so I moved the logic that checks the `exclude` property to `_asdict` – before the function calls itself recursively on the items of the excluded field.

With a dataclass that contains a few simple fields as well as an excluded `List[int]` of length 10 000 000, this reduced the runtime from over 30 seconds to a small fraction of a second on my machine.


george-zubrienko:
@stafak hello :) could you resolve conflicts and add unit tests to back your statement?

PR #249, opened 2020-09-30
fix: vendor-in code from stringcase

### LINKED ISSUE #236: Is the dependency on stringcase really needed?
Hi!

I'm considering packaging this library in Debian for another project, and looking at the dependencies, I'm not sure I understand the need to depend on [stringcase](https://github.com/okunishinishi/python-stringcase).

The stringcase project doesn't seem very well maintained and that makes it harder for me to package it:
* no releases
* no git tags
* last release on PyPi was 3 years ago

Wouldn't it be possible to implement the 20 lines of python `dataclasses-json` needs directly in `api.py`? `stringcase` is also licensed under the MIT and it would be possible to simply copy the needed code.

If needed, I'd be happy to submit a PR myself.

baldurmen:
Closes #236.

The python-stringcase dependency this package has is very badly maintained:
  * There are no releases.
  * There are no git tags.
  * The last release on PyPi was 3 years ago, although internal version
    numbers have been bumped since then.
  * The project seems somewhat dormant, if not abandonned.

Having it as a dependency is thus a liability and makes this library hard to package in Linux distributions, as packaging badly maintained libraries is no fun.

Since the code used is actually very simple, it makes sense to vendor it in and not depend on it anymore.

Note that I am not the author of the majority of the code in this patch and mainly did some glue work. Credit and copyright attributions (under the MIT) should be attributed to the original author, Taka Okunishi <okunishinishi@gmail.com>. I tried to make that clear by adding a copyright header to the new stringcase.py file.

The original code can be found at: https://raw.githubusercontent.com/okunishinishi/python-stringcase/master/stringcase.py

I ran the testsuite and `flake8` and everything passes. Happy to make any modifications required :)

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -4,9 +4,8 @@ from enum import Enum
 from typing import (Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar,
                     Union)
 
-from stringcase import (camelcase, pascalcase, snakecase,
-                        spinalcase)  # type: ignore
-
+from dataclasses_json.stringcase import (camelcase, pascalcase, snakecase,
+                                         spinalcase)  # type: ignore
 from dataclasses_json.cfg import config
 from dataclasses_json.core import (Json, _ExtendedEncoder, _asdict,
                                    _decode_dataclass)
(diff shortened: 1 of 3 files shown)


baldurmen:
Ah, forgot to say, I did this while working on [packaging dataclasses-json in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=971338). Merging and releasing this wouldl make my job much much easier :)

baldurmen:
Hello! Could anyone look at this patch? I'd very much like to stop carrying it in Debian :)

Cheers, and thanks for this lib.

PR #251, opened 2020-10-06
handle `dict` when the key is a `tuple`, fixes #250

### LINKED ISSUE #250: `to_dict` fails when a dict key is a tuple
```
from typing import List, Optional, Dict, cast, Tuple, Union

@dataclass
class State(DataClassJsonMixin):

    foo: Dict[Tuple[str, str], str]

state = State(
    foo = {('abc', 'cde'): 'efg'}
)
state.to_dict()
```
```
  File "/work/ve/lib/python3.7/site-packages/dataclasses_json/api.py", line 88, in to_dict
    return _asdict(self, encode_json=encode_json)
  File "/work/ve/lib/python3.7/site-packages/dataclasses_json/core.py", line 321, in _asdict
    value = _asdict(getattr(obj, field.name), encode_json=encode_json)
  File "/work/ve/lib/python3.7/site-packages/dataclasses_json/core.py", line 331, in _asdict
    obj.items())
TypeError: unhashable type: 'list'
```

agroszer:


matt035343:
Thanks for your time. Using tuples for dict keys is not JSON serializable. Therefore, we cannot go through with this PR.

PR #255, opened 2020-10-14
Allow the global config dictionary keys to also be Optional[type]

sumnerevans:
When the decoder is for an `Optional` type, then the mypy type checker fails since the encoders/decoders are of type `Dict[type, Callable]`.

Example that fails:
```
dataclasses_json.cfg.global_config.decoders[Path] = Path
dataclasses_json.cfg.global_config.decoders[Optional[Path]] = (
    lambda p: Path(p) if p else None
)
```

mypy gives the following error in this case:
```
error: Invalid index type "object" for "Dict[type, Callable[..., Any]]"; expected type "type"
```

This PR fixes this issue by allowing for `Optional[type]` as well as `type`.


george-zubrienko:
@matt035343 take a quick look here? Do we need units for this or can merge?

matt035343:
Always nice with tests, but I am not sure how to test this 🤷‍♂️

george-zubrienko:
> Always nice with tests, but I am not sure how to test this 🤷‍♂️

Merge and wait for annoyed users to show up 😂

george-zubrienko (approved):

PR #262, opened 2020-11-12
Fix typing when inheriting from DataClassJsonMixin

OdinTech3:
PR fixes the below Mypy Errors: 

* `Incompatible types in assignment (expression has type "Dict[Any, Any]", base class "DataClassJsonMixin" defined the type as "None")`

* `Argument "letter_case" to "config" has incompatible type "LetterCase"; expected "Optional[Callable[[str], str]]"`

When doing something like:
```python
class MyDataClass(DataClassJsonMixin):
    dataclass_json_config = config(letter_case=LetterCase.CAMEL, undefined=Undefined.EXCLUDE)["dataclasses_json"]
```


OdinTech3:
Gonna try and fix the circular import issue as soon as I can.

USSX-Hares:
- LetterCase parameter type is reported as #286 and fixed as #343
 - `dataclass_json_config`, if I understand correctly, has no type annotation in order to not interfere the DataClasses field logic. @lidatong, am I correct?

dataclasses_json/cfg.py
```diff
@@ -41,30 +39,32 @@ def __init__(self):
...
-        lib_metadata['encoder'] = encoder
+        lib_metadata["encoder"] = encoder
 
     if decoder is not None:
-        lib_metadata['decoder'] = decoder
+        lib_metadata["decoder"] = decoder
 
     if mm_field is not None:
-        lib_metadata['mm_field'] = mm_field
+        lib_metadata["mm_field"] = mm_field
```
USSX-Hares, line 67: Here and below: why do you change the single quotes for the double ones?

dataclasses_json/cfg.py
```diff
@@ -41,30 +39,32 @@ def __init__(self):
...
+    # TODO: these can be typed more precisely
+    # Specifically, a Callable[A, B], where `B` is bound as a JSON type
+    encoder: Callable = None,
+    decoder: Callable = None,
+    mm_field: MarshmallowField = None,
+    letter_case: Union[Callable[[str], str], 'LetterCase'] = None,
+    undefined: Optional[Union[str, Undefined]] = None,
+    field_name: str = None,
+    exclude: Optional[Callable[[str, T], bool]] = None,
+) -> Dict[str, dict]:
```
USSX-Hares, line 54: Why do you change the existing lines indentation?

OdinTech3: I think it was because I formatted w/ black.

dataclasses_json/cfg.py
OdinTech3: I think it was because I formatted w/ black.

george-zubrienko:
I'm closing this one as stale and having unresolved comments. Feel free to open an issue and re-create the PR if you come by ;)

PR #269, opened 2020-12-07
Allow nested dataclasses-json objects 

vribic:


vribic:
#255  Proposed solution for handling nested dataclasses-json objects by keeping a local collection of classes inheriting or being register to DataClassJsonMixin.

PR #279, opened 2021-02-16
Add `tuple` to the list of fields.

gyscos:
Without it, getting a schema from a type including `typing.Tuple` returns a warning (as `typing.Tuple.__origin__` is `tuple`).


lidatong (approved):
Thanks for the contribution!

PR #285, opened 2021-03-06
Add MANIFEST.in

### LINKED ISSUE #280: LICENSE file missing in pypi.org release
The tar.gz file from pypi.org

https://pypi.org/project/dataclasses-json/#files

does not contain the LICENSE file. The one from here does:

https://github.com/lidatong/dataclasses-json/releases/tag/v0.5.2

Is this on purpose?

The LICENSE file is nice to have when using your library in Yocto builds, because Yocto checks the licenses of included projects.

fabaff:
Ensure that the license file is part of the source tarball published on PyPI.

It would be nice if you could create a new release after merging because some distributions require to ship the license file, e. g. Fedora. Thanks

Fixes #280

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE


lidatong (approved):
Yes I neglected this -- thanks for fixing!

Issue #294, opened 2021-05-06
TypeError on from_dict on dataclass with nested dict and tuple


christian-steinmeyer:
When creating a dataclass with a nested dict and a tuple, `from_dict()` does not work.

For a minimal working example:
```python
from typing import Dict, Tuple
from dataclasses import dataclass
from dataclasses_json import DataClassJsonMixin

@dataclass
class TestClass(DataClassJsonMixin):
    nested_dict: Dict[str, Dict[Tuple[str], int]]

my_test = TestClass.from_dict({"nested_dict": {"a":{("b", "c"): 1, "c": 2}, "d": {}}})
```

This yields:
```bash
Traceback (most recent call last):
  File "[PROJECT_PATH]python3.8/site-packages/dataclasses_json/core.py", line 263, in _decode_generic
    res = _get_type_cons(type_)(xs)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/typing.py", line 727, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
TypeError: Type Tuple cannot be instantiated; use tuple() instead
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "[PROJECT_PATH]python3.8/site-packages/dataclasses_json/core.py", line 263, in _decode_generic
    res = _get_type_cons(type_)(xs)
  File "[PROJECT_PATH]python3.8/site-packages/dataclasses_json/core.py", line 309, in <genexpr>
    items = (_decode_generic(type_arg, x, infer_missing) for x in xs)
  File "[PROJECT_PATH]python3.8/site-packages/dataclasses_json/core.py", line 265, in _decode_generic
    res = type_(xs)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/typing.py", line 727, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
TypeError: Type Dict cannot be instantiated; use dict() instead
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "[PROJECT_PATH]python3.8/site-packages/IPython/core/interactiveshell.py", line 3437, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-15-a734de727315>", line 1, in <module>
[...]

PR #297, opened 2021-05-18
fix: decoding nested dicts with tuple keys

pearceeverydaylabs:
suggested fix for https://github.com/lidatong/dataclasses-json/issues/294

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -17,7 +17,7 @@ from uuid import UUID
 from typing_inspect import is_union_type  # type: ignore
 
 from dataclasses_json import cfg
-from dataclasses_json.utils import (_get_type_cons,
+from dataclasses_json.utils import (_get_type_cons, _get_type_origin,
                                     _handle_undefined_parameters_safe,
                                     _is_collection, _is_mapping, _is_new_type,
                                     _is_optional, _isinstance_safe,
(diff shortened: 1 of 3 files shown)


christian-steinmeyer (commented):
Should do the trick!

pearceeverydaylabs:
@lidatong could you review, please?

dataclasses_json/core.py
```diff
@@ -285,11 +285,20 @@ def _decode_dict_keys(key_type, xs, infer_missing):
...
+    if key_type is None or key_type == Any:
+        decode_function = key_type = (lambda x: x)
+    # handle a nested python dict that has tuples for keys. E.g. for
+    # Dict[Tuple[int], int], key_type would be typing.Tuple[int], but
+    # decode_function would be tuple, so map() doesn't break
+    # the key_type would be typing.Tuple[int]
+    elif _get_type_origin(key_type) == tuple:
+        decode_function = tuple
+        key_type = key_type
+
```
lidatong, line 300: what would happen in the `else` case?

lidatong: never mind i see the original logic was effectively `key_type` = `key_type`

lidatong (approved):
thanks for the contribution! appreciate it

pearceeverydaylabs:
@lidatong thanks for approving! I pushed up a fix for python 3.6 that I got working locally in a 3.6 venv. Could you please approve re-running workflows since I don't have permission? :) 
Screen Shot 2021-06-04 at 12 20 10 AM

lidatong:
Sure -- and looks like checks are passing now. Thanks again for the contribution!

Issue #300, opened 2021-07-07
Python 3.6: Unsupported Generic types raise an exception during decoding


USSX-Hares:
### Explanation
Python 3.6 uses a different model for ABCs and Generics, and unsupscripted Generics *have* attribute `__args__`, but it is `None`.
So, if you are decoding a complex model and the dataclass field type is non-dataclass unsupported generic, Python 3.6 would raise an exception in function `dataclasses_json.utils._hasargs` which tries to search for `arg` in `type_.__args__` which is `None` and thus not possible.

### Code Snippet
```python
import json
from typing import *

from dataclasses import dataclass
from dataclasses_json import DataClassJsonMixin

T = TypeVar('T')
class ClassA(Generic[T]):
    value: T

class ClassB:
    a: ClassA
    x: str


@dataclass
class ClassAImpl(ClassA, DataClassJsonMixin):
    value: T

@dataclass
class ClassBImpl(ClassB, DataClassJsonMixin):
    a: ClassA
    x: str


def main():
    j = \
    """
    {
        "a": { "value": 22 },
        "x": "should be parsed"
    }
    """
    
    data = json.loads(j)
    b: ClassB = ClassBImpl.from_dict(data)   # Here b.a is loaded but not interpreted as ClassA / ClassAImpl
    b.a = ClassAImpl.from_dict(data['a'])    # So, let's load ClassA from `data['a']` and store it as `b.a`
    print(b)
    
    return 0


if (__name__ == '__main__'):
    exit_code = main()
    exit(exit_code)
```

#### Python 3.6
```
> C:\git\gitlab\tools\openapi-parser\.venv-36\Scripts\python.exe C:/git/gitlab/tools/openapi-parser/dcj-with-py36-generic.py
Traceback (most recent call last):
  File "C:/git/gitlab/tools/openapi-parser/dcj-with-py36-generic.py", line 44, in <module>
    exit_code = main()
  File "C:/git/gitlab/tools/openapi-parser/dcj-with-py36-generic.py", line 36, in main
    b: ClassB = ClassBImpl.from_dict(data)   # Here b.a is loaded but not interpreted as ClassA / ClassAImpl
  File "C:\git\gitlab\tools\openapi-parser\.venv-36\lib\site-packages\dataclasses_json\api.py", line 83, in from_dict
    return _decode_dataclass(cls, kvs, infer_missing)
[...]

PR #301, opened 2021-07-07
Fix unsupported generics for Python 3.6

USSX-Hares:
### Changes:
 - Added tests to detect #300 (two fail on Python 3.6)
   * Unsupscripted Generics
   * Supscripted Generics
   * Parameterized Generics in Unsupscritped Generic class
   * Parameterized Generics in Supscritped Generic class
 - Updated dev dependencies to install `types-dataclasses` for Python 3.6 (to fix Annotations tests)
 - Updated `_decode_dataclass` to use `_isinstance_safe` instead of simple `isinstance`
 - Added TypeError handler for `_hasargs` function

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -128,7 +128,7 @@ def _decode_letter_case_overrides(field_names, overrides):
 
 
 def _decode_dataclass(cls, kvs, infer_missing):
-    if isinstance(kvs, cls):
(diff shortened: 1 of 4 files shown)


lidatong:
Thank you for the reporting the issue and the accompanying fix!

lidatong (approved):

PR #312, opened 2021-08-31
bug: override encoding before _asdict

### LINKED ISSUE #311: re-implementation of `_asdict` fails to parse `DataFrame`
Given a class like
```python
from dataclasses_json.cfg import config
import pandas
from pandas import DataFrame
import numpy

@dataclass_json
@dataclass
class SomeClass:
    name: str
    dataframe: DataFrame = field(
        metadata=config(
            field_name="someOverriddenName",
            decoder = lambda x: DataFrame(data=numpy.array(x))
        )
    )

    def __post_init__(self):
        self.dataframe = pandas.DataFrame(
            data = numpy.array(self.dataframe).reshape(
                2, 2
            ),
            columns=['some', 'names'],
            index=['some', 'names'],
        )
		```
then creating an instance: 
```python
x = SomeClass.from_dict({'name': 'a name', 'someOverriddenName': [1, 0, 0, 1]})
x.dataframe
# returns the expected DataFrame
```

then calling `x.to_json()` or `x.to_dict()` only returns the row names of the DataFrame:

```python
x.to_json()
# '{"name": "a name", "someOverriddenName": ["some", "names"]}'
x.to_dict()
# {'name': 'a name', 'someOverriddenName': ['some', 'names']}
```

This appears to be due to a bug in l. 348 of `core.py`:
```python
def _asdict(obj, encode_json=False):
	# ...
	return list(_asdict(v, encode_json=encode_json) for v in obj)
	# ...
```
which tries to call `_asdict` on only the column names of the `DataFrame`.

What I'd really like to do here is override the `config` with a custom `encoder` function, but `_asdict` is called _before_ the field is passed to the `encoder`, so my custom function only recieves the `list` of row names.

gegnew:
Would love your feedback on this solution

Closes #311

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -335,7 +335,10 @@ def _asdict(obj, encode_json=False):
             if overrides[field.name].encoder:
                 value = getattr(obj, field.name)
             else:
-                value = _asdict(getattr(obj, field.name), encode_json=encode_json)
+                value = _asdict(
+                    getattr(obj, field.name),
+                    encode_json=encode_json
+                )
             result.append((field.name, value))
 
         result = _handle_undefined_parameters_safe(cls=obj, kvs=dict(result),
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -330,8 +330,12 @@ def _asdict(obj, encode_json=False):
     """
     if _is_dataclass_instance(obj):
         result = []
+        overrides = _user_overrides_or_exts(obj)
         for field in fields(obj):
-            value = _asdict(getattr(obj, field.name), encode_json=encode_json)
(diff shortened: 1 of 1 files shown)


gegnew:
fixed linting

gegnew:
bump; all tests should pass. If this minor change works for you, I'd love to have the core package, and not my fork, as a dependency

gegnew:
whoops, sorry, I ran `black` to fix a linting issue and it changed way more than I thought

gegnew:
fixed linting _and_ undid all the formatting with `black`

lidatong (approved):
sorry for the delay and thanks for addressing the CI issues! merging

gegnew:
Thanks!

PR #346, opened 2022-03-22
#341: Fixes for type-parameters without arguments

### LINKED ISSUE #341: `typing` module types without parameters broken when using Python 3.9
When using types in the built-in `typing` module, if a type supports additional parameters and those parameters are not used, errors result when using Python 3.9

Code:
```python
from typing import List

@dataclass
class Parent(DataClassJsonMixin):
    children: List


if __name__ == '__main__':
    p = Parent.from_dict({
        "children": ['hi']
    })
    print(p.to_json())
```

Output when running dataclasses-json 0.5.6 and Python 3.9:
```
  File "/Users/daniel.letscher/code/playground/dataclass.py", line 13, in <module>
    p = Parent.from_dict({
  File "/Users/daniel.letscher/code/playground/venv39/lib/python3.9/site-packages/dataclasses_json/api.py", line 82, in from_dict
  File "/Users/daniel.letscher/code/playground/venv39/lib/python3.9/site-packages/dataclasses_json/core.py", line 201, in _decode_dataclass
  File "/Users/daniel.letscher/code/playground/venv39/lib/python3.9/site-packages/dataclasses_json/core.py", line 258, in _decode_generic
    xs = _decode_items(type_.__args__[0], value, infer_missing)
  File "/Users/daniel.letscher/.pyenv/versions/3.9.7/lib/python3.9/typing.py", line 706, in __getattr__
    raise AttributeError(attr)
AttributeError: __args__
```

Output when running dataclasses-json 0.5.6 and Python 3.8:
```
{"children": ["hi"]}
```

USSX-Hares:
Closes #341

### Changes:
 * Updated `tests.entities` and `tests.test_collections` to include alternative forms of the same collection types
   (like `builtins.list` ("builtin") and `typing.List` (~~"generic"~~ "unbound") in addition to the existing `typing.List[TypeA]`)
 * Added new internal helper class ~~`core._NoArgs`~~ `utils._NoArgs` and its singleton instance ~~`core._NO_ARGS`~~ `utils._NO_ARGS`
 * Added new methods ~~`core._extract_args()`~~ `utils._get_type_args()` and ~~`core._extract_type_parameter()`~~ `utils._get_type_arg_param()` which return either `tp.__args__` value or `_NO_ARGS` constant
 * All "unsafe" usages of `tp.__args__` replaced with the new methods
 * Added `TypeVar` special case check for `core._decode_dict_keys()`
   - *Note:* This is a special case for Python 3.7 and Python 3.8.
     By some reason, "unbound" dicts are counted as having key type parameter to be `TypeVar('KT')`
 * Added `tp.__origin__ == None` special case for `utils._get_type_origin()`
   - *Note:* This is a special case for Python 3.6 exclusively.
     By some reason, *some* "unbound" types (`typing.Set`, `typing.Tuple`, and `typing.FrozenSet`) are considered to have this field with `None` rather than generating an `AttributeError`
   - Same for `tp.__args__` and dicts

### Status
 * Without the fix, new tests fail on all versions of Python, but fail differently (see report in the comments of #341)
 * By now, there are no failures related to the missing `tp.__args__`:
   + [x] Python 3.6: ~~Decodes "unbound" (but not "builtin") sets, tuples, and frozensets as lists rather than required types~~
   + [x] Python 3.7-3.8: ~~Fails to decode "unbound" and "builtin" dicts by trying to instantiate uninstantiatable `typing.Dict`.~~
      Was actually caused by the strange structure of dict class nesting in specifically these Python versions, see note above
   + [x] No longer fails on Python 3.9+
   + [x] *(New)* Python 3.6: ~~Unable to decode "unbound" `typing.Dict` and `dict` keys on Python 3.6 exclusively~~

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -329,8 +329,10 @@ def _decode_dict_keys(key_type, xs, infer_missing):
     decode_function = key_type
     # handle NoneType keys... it's weird to type a Dict as NoneType keys
     # but it's valid...
-    # Issue #341 and MR #346: This is a special case for Python 3.7 and Python 3.8.
-    # By some reason, "unbound" dicts are counted as having key type parameter to be TypeVar('KT')
+    # Issue #341 and PR #346:
+    #   This is a special case for Python 3.7 and Python 3.8.
+    #   By some reason, "unbound" dicts are counted
+    #   as having key type parameter to be TypeVar('KT')
     if key_type is None or key_type == Any or isinstance(key_type, TypeVar):
         decode_function = key_type = (lambda x: x)
     # handle a nested python dict that has tuples for keys. E.g. for
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -329,6 +329,8 @@ def _decode_dict_keys(key_type, xs, infer_missing):
     decode_function = key_type
     # handle NoneType keys... it's weird to type a Dict as NoneType keys
     # but it's valid...
+    # Issue #341 and MR #346: This is a special case for Python 3.7 and Python 3.8.
+    # By some reason, "unbound" dicts are counted as having key type parameter to be TypeVar('KT')
     if key_type is None or key_type == Any or isinstance(key_type, TypeVar):
         decode_function = key_type = (lambda x: x)
     # handle a nested python dict that has tuples for keys. E.g. for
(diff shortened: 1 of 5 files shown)


UberMeatShield:
This seems like a pretty solid thing to get into the project so it is working in 3.10 as well :)

lidatong:
Yes I'm sorry I've been MIA... I'll take a look soon!

lidatong (approved):

cprasad1:
@lidatong do you have plans to release a new version with this patch in? It will be great if we can get this bug fixed

PR #352, opened 2022-04-12
Add support for encoding numpy integer and floating point types

pvanheus:
Feel free to ignore if this is too much of an edge case for your module, but I found a need to serialise dataclasses to JSON that had NumPy types (int64 etc) in them. This PR adds support for encoding NumPy types, but only if the numpy module is available.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -16,6 +16,12 @@ from uuid import UUID
 
 from typing_inspect import is_union_type  # type: ignore
 
+try:
+    import numpy as np
+    np_available = True
+except ImportError:
+    np_available = False
+
 from dataclasses_json import cfg
 from dataclasses_json.utils import (_get_type_cons, _get_type_origin,
                                     _handle_undefined_parameters_safe,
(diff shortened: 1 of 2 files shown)


sebastian-ruiz:
It would be nice to handle numpy arrays as well. Something like:
```python
elif np_available and _isinstance_safe(o, np.ndarray):
    result = o.tolist()
```

pvanheus:
> It would be nice to handle numpy arrays as well. Something like:
> 
> ```python
> elif np_available and _isinstance_safe(o, np.ndarray):
>     result = o.tolist()
> ```
Thanks, I have belatedly incorporated your suggestion, simplified the code and added a test for encoding to JSON,

george-zubrienko:
I wonder why not use encoder/decoder instead? I've seen people do this with fields being pandas Dataframes

matt035343:
Thanks for your PR. Imo, I believe dataclasses-json should only support native types out of the box while 3rd-party types like numpy can be de/encoded with custom encoders.

george-zubrienko:
I'd say if the author added instead an example for docs with custom encoder/decoder for numpy, that would have helped many people. I'll wait a bit, if no reaction follows, I'll close this one and update docs with examples for both numpy and pandas.

matt035343:
Closing this in favor of #470

Issue #358, opened 2022-05-03
Different field_name on different object returned


Dvergatal:
I am currently struggling with a difficult problem to solve. because what i want to do is to prepare a data class for azure cloud to i.e. [update device attestation](https://docs.microsoft.com/en-us/rest/api/iotcentral/1.2-previewdataplane/devices/update-attestation).

The problem which I'm facing is that a sample request for i.e. Symmetric Key looks like this:
```json
{
  "type": "symmetricKey",
  "symmetricKey": {
    "primaryKey": "<primary key>",
    "secondaryKey": "<secondary key>"
  }
}
```
as you can see in another json code block for i.e. TPM:
```json
{
  "type": "tpm",
  "tpm": {
    "endorsementKey": "<endorsement key>"
  }
}
```
you can see that the second field changes according to the name of class. Now I have these classes written and have a factory function returning, specified by this first type variable, object. The problem which I'm facing is that the name of this object in json is the name of variable not the name of the returned object type. I was thinking of using this field_name parameter from the field module, in order to change it's name but it's not possible, or maybe i dunno how to do that, to change it depending on the returned object type.

I would be pleased if somebody might have a solution for this use case.

PR #359, opened 2022-05-05
Add support for setting field name from class name.

Dvergatal:
Hi all.
I have added support which i have been missing on issue #358. I hope that we could discuss about my solution and if something is wrong in your opinion please let me know.


lidatong (approved):

Dvergatal:
@lidatong I am not allowed to merge it.

george-zubrienko (commented):
@Dvergatal I can merge it right away, but would you be so kind to add a unit test for this change?

PR #363, opened 2022-06-27
Fix deserialization for typing.Counter

n1chre:
The issue here is that typing.Counter deserialization ends ups in the generic deserialization block. And because Counter is a mapping, we try to get key and value type of it. But it fails, and we shouldn't really deserialize a counter that way. This PR fixes it and deserializes it properly

Example test:
```py3
$ cat test.py
import typing
import collections

@dataclass_json
@dataclass
class C:
    c: typing.Counter[str]

c = C(collections.Counter('aba'))
print(c.to_json())
print(C.from_json('{"c":{"a":2,"b":1}}'))
```

Before:
```bash
$ py3 test.py
{"c": {"a": 2, "b": 1}}
  File "/Users/hrenic/test.py", line 13, in <module>
    print(C.from_json('{"c":{"a":2,"b":1}}'))
  File "/opt/homebrew/lib/python3.9/site-packages/dataclasses_json-0.5.7-py3.9.egg/dataclasses_json/api.py", line 65, in from_json
  File "/opt/homebrew/lib/python3.9/site-packages/dataclasses_json-0.5.7-py3.9.egg/dataclasses_json/api.py", line 72, in from_dict
  File "/opt/homebrew/lib/python3.9/site-packages/dataclasses_json-0.5.7-py3.9.egg/dataclasses_json/core.py", line 202, in _decode_dataclass
  File "/opt/homebrew/lib/python3.9/site-packages/dataclasses_json-0.5.7-py3.9.egg/dataclasses_json/core.py", line 255, in _decode_generic
ValueError: not enough values to unpack (expected 2, got 1)
```

After: 
```bash
$ py3 test.py
{"c": {"a": 2, "b": 1}}
C(c=Counter({'a': 2, 'b': 1}))
```

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -11,7 +11,8 @@ from dataclasses import (MISSING,
 from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
-from typing import Any, Collection, Mapping, Union, get_type_hints, Tuple
+from typing import (Any, Collection, Counter, Mapping,
+                    Union, get_type_hints, Tuple)
 from uuid import UUID
 
 from typing_inspect import is_union_type  # type: ignore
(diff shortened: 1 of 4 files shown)


george-zubrienko:
'typing.Counter' is a deprecated alias to collections class, according to docs https://docs.python.org/3/library/collections.html#collections.Counter

Does that change anything for us here?

matt035343:
@n1chre I tried out your solution, but it seems like there are is a bug when combining with the changes of #435. I have fixed this in a new PR #471. If you don't mind, I will go ahead and have #471 approved and merged instead. Let me know what you think.

matt035343:
Closed in favor of #471

Issue #336, opened 2022-02-15
Exclude function type signature does not match usage


artificial-aidan:
`exlude` accepts a `Callable[[str, Any], bool]` [here](https://github.com/lidatong/dataclasses-json/blob/a5fd1b53b4e9c192a2ecfd851502d056309c36c8/dataclasses_json/cfg.py#L53)

But when it is invoked only the value is passed

https://github.com/lidatong/dataclasses-json/blob/a5fd1b53b4e9c192a2ecfd851502d056309c36c8/dataclasses_json/core.py#L103

artificial-aidan:
https://github.com/lidatong/dataclasses-json/pull/231

Fixed here

@lidatong

PR #368, opened 2022-07-12
Fix config(exclude=…) type annotations

andersk:
* The `exclude` callback takes one argument, not two.
* Objects of type `Exclude` are not accepted. `ALWAYS` and `NEVER` have type `Callable`, not `Exclude`.
* Class variables like `Exclude.ALWAYS`, `Exclude.NEVER` are not allowed to be generic, and they don’t need to be, since `Callable[[object], bool]` is convertible to `Callable[[T], bool]`.

### DIFF
--- a/dataclasses_json/cfg.py
+++ b/dataclasses_json/cfg.py
@@ -16,8 +16,8 @@ class Exclude:
     Pre-defined constants for exclusion. By default, fields are configured to
     be included.
     """
-    ALWAYS: Callable[[T], bool] = lambda _: True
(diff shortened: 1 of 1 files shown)


andersk:
Apparently I’m not the first to submit this.

- #231 
- #336
- #360

george-zubrienko:
hi @andersk could you please update your PR so we can review and merge :)

andersk:
@george-zubrienko Rebased.

george-zubrienko (approved):

PR #374, opened 2022-09-27
Update usage of mypy.main in test following updates to mypy package

rpmcginty:
Last month, a **mypy** change was released that made all arguments for the `mypy.main` method **keyword** only. They also removed an unused argument. 

https://github.com/python/mypy/pull/13399

The alternative to this change is to freeze your version of `mypy`

### DIFF
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -105,7 +105,7 @@ class TestAnnotations:
         try:
             # mypy.main uses sys.stdout for printing
             # We override it to catch error messages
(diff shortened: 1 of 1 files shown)


lidatong (approved):

lidatong:
@rpmcginty thanks for this! `mypy_main` probably changed again probably between the time you submitted this PR and now... apologies again for that. i skipped the test to get checks passing on `master`, but feel free to remove that

i suppose we could pin `mypy` but to be honest i don't know enough about the typing ecosystem but i'm concerned about weird unlucky incompatibilities for users on different versions...

Issue #348, opened 2022-04-04
Add support for `str` subclasses 


rpmcginty:
### Problem
as of v0.5.7, any subclassed `str` classes are instantiated using the default `str` class. 

#### Example
```python
class MyString(str):
    ...
    def my_useless_odd_number_detector(self) -> bool:
        return len(self) % 2 != 0

@dataclass_json
@dataclass
class MyDC:
    my_str: MyString

    def __post_init__(self):
        print(f"IS it odd? -> {self.my_str.my_useless_odd_number_detector()}")

# Error!
MyDC.from_dict(dict(my_str="123"))
```

To work around this, you must override the `decoder`. But this must be done in every dataclass that references that type. Alternatively, we can add this decoder to the global config. but this can be cumbersome if you are subclassing str many times. 

For context, I have a subclassed `str` into `ValidatedStr` which provides validation (regex, length constraints) against the string. I vend this `ValidatedStr` class for usage across other packages. I do not want to require others using this subclass to have to explicitly specify the decoder every time or update a global config.   

### Feature Request

Modify the `decode_dataclass` logic to further distinguish between custom subclassed `str` classes.

george-zubrienko:
I think you should use composition instead of inheritance for this use case - would make life much easier. As for the library behaviour, I agree this code should not raise

matt035343:
This seems to be fixed as of version 0.5.9. Please reopen if the issue persists.

PR #375, opened 2022-09-27
Add support for enhanced builtin instantiation

rpmcginty:
## What's in this change?
* Adding improved instantiation logic for `str`, `int`, `float` in dataclass decoder. 
* Fixing bugs in test code 

### Why do we need improved instantiation logic for these builtins?

There are a few reasons why instantiating using the intended type when they are one of the three is beneficial. 

1. **Support for custom `str` classes is covered.** Previously this was not supported as `str` is treated special in the decode logic. This is further documented here: https://github.com/lidatong/dataclasses-json/issues/348
2. **Improved support for conversion from input X -> `int`|`float`**. Previously, values that represented numeric values (either stringified numeric values or `Decimal` objects) were not converted properly into the intended field type. So a dataclass field specified as an `int` would not decode `"42"` or `decimal.Decimal("42")` as `42`. This differs from *marshmallow* decode behavior which properly converts to the integer value.  

## Testing
Added tests to demonstrate new usage

Note that this will fail tests until this PR is merged: https://github.com/lidatong/dataclasses-json/pull/374

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -230,6 +230,10 @@ def _support_extended_types(field_type, field_value):
         res = (field_value
                if isinstance(field_value, UUID)
                else UUID(field_value))
+    elif _issubclass_safe(field_type, (int, float, str, bool)):
+        res = (field_value
+               if isinstance(field_value, field_type)
+               else field_type(field_value))
     else:
         res = field_value
     return res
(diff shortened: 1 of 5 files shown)


lidatong (approved):

george-zubrienko:
@rpmcginty could you please sync your branch so CI runs and then we can merge :)

rpmcginty:
@george-zubrienko done!

george-zubrienko:
> @george-zubrienko done!

Thank you!If there is anything else pending merge, just sync the PR and ping me in comments :)

Issue #380, opened 2022-10-11
Drop marshmallow_enum dependency


deuxnids:
Hello,

With the release of marshmallow 3.18.0 the Enum type is supported (https://github.com/marshmallow-code/marshmallow/blob/dev/CHANGELOG.rst#3180-2022-09-15).  The repository for marshmallow_enum will be closed (https://github.com/justanr/marshmallow_enum/issues/51).

It would be nice to remove the dependency to marshmallow_enum in https://github.com/lidatong/dataclasses-json/blob/master/setup.py#L21, since marshmallow_enum still depends on 'marshmallow>=2.0.0' which suffers from CVE-2018-17175 vulnerability.

What do you think?

baldurmen:
I'm also keen for this to happen, as this would let me drop the [python-marshmallow-enum](https://tracker.debian.org/pkg/python-marshmallow-enum) package I maintain in Debian for the sole purpose of packaging dataclasses-json :)

PR #384, opened 2022-10-23
Drop marshmallow-enum dependency

a-was:
According to #380 this PR drops [marshmallow-enum](https://github.com/justanr/marshmallow_enum) dependency, which is now an archive. 
Enum type is supported since marshmallow 3.18.0.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,8 +9,7 @@ mypy = "*"
 pytest = "6.2.3"
 
 [packages]
-marshmallow = "*"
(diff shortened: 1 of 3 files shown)


george-zubrienko:
@a-was could you update your PR please, so we can merge? :)

a-was:
@george-zubrienko done, check if everything is okay

george-zubrienko:
Closes #380

george-zubrienko (approved):

PR #394, opened 2022-12-05
Fixes recursive dict encoding

### LINKED ISSUE #391: `to_dict(encode_json=True)` is not recursive
I expect `to_dict(encode_json=True)` to be recursive, however the code below evaluates to:
```
{'engine': 'fast', 'wheels': [<Spec.fast: 'fast'>, <Spec.slow: 'slow'>]}
```
The first level of Enums evaluate to strings as expected, but the list is not JSON encoded. I expected it to be:
```
{'engine': 'fast', 'wheels': ['fast', 'slow']}
```

```
from enum import Enum
from typing import List

class Spec(Enum):
    fast = 'fast'
    slow = 'slow'

@dataclass
class Car(DataClassJsonMixin):
    engine: Spec
    wheels: List[Spec]

car = Car(engine=Spec.fast, wheels=[Spec.fast, Spec.slow])

car.to_dict(encode_json=True)
```

My current work-around is:
```
import json

json.loads(car.to_json())
```
However, this is not as nice

matt035343:
Fixes #391, please refer to this issue for a detailed description of what is fixed in this pull request.

I implemented both a fix to the encoding and added tests that will capture the behaviour.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -93,7 +93,12 @@ def _user_overrides_or_exts(cls):
 
 def _encode_json_type(value, default=_ExtendedEncoder().default):
     if isinstance(value, Json.__args__):  # type: ignore
-        return value
+        if isinstance(value, list):
+            return [_encode_json_type(i) for i in value]
+        elif isinstance(value, dict):
+            return {k: _encode_json_type(v) for k, v in value.items()}
+        else:
+            return value
     return default(value)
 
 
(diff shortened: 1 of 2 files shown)


AbsurdlySuspicious:
Wouldn't this still fail on dicts or lists with more than one level of nesting? I believe this type check should be done in `_encode_json_type` to be properly recursive

matt035343:
@AbsurdlySuspicious

> Wouldn't this still fail on dicts and lists with more than one level of nesting? I believe this type check should be done in `_encode_json_type` to be properly recursive

You are absolutely right. I have now fixed it and changed the test to catch this case as well.

george-zubrienko:
@matt035343 could you sync your branch with master, so CI runs again w/o py3.6?

george-zubrienko (approved):

Issue #395, opened 2022-12-21
Generated schema doesn't enforce required fields


deansg:
Given some class:
```
@dataclass_json(letter_case=LetterCase.CAMEL, undefined=Undefined.RAISE)
@dataclass
class MyClass:
    required_field: str
    other_field: str = None
```
I expect the schema generated using MyClass.schema() to mark the first field (`required_field`) which has no default value as required. However, this doesn't happen, and if I try to use the schema to validate an object without a value for the required field, we get a `KeyError` instead of a proper validation error.

PR #397, opened 2022-12-23
#395: fields with no default value/factory should be marked as required when generating schema

deansg:
When generating a marshmallow schema for a dataclass, fields that have no default value or default factory should be marked as required, so that when deserializing an object into the dataclass, a missing value for the field will result in a proper Validation error instead of a KeyError.

### DIFF
--- a/dataclasses_json/mm.py
+++ b/dataclasses_json/mm.py
@@ -282,18 +282,19 @@ def schema(cls, mixin, infer_missing):
     # TODO check the undefined parameters and add the proper schema action
     #  https://marshmallow.readthedocs.io/en/stable/quickstart.html
     for field in dc_fields(cls):
-        metadata = (field.metadata or {}).get('dataclasses_json', {})
         metadata = overrides[field.name]
         if metadata.mm_field is not None:
             schema[field.name] = metadata.mm_field
         else:
             type_ = field.type
-            options = {}
+            options: typing.Dict[str, any] = {}
             missing_key = 'missing' if infer_missing else 'default'
             if field.default is not MISSING:
                 options[missing_key] = field.default
             elif field.default_factory is not MISSING:
                 options[missing_key] = field.default_factory
+            else:
+                options['required'] = True
 
             if options.get(missing_key, ...) is None:
                 options['allow_none'] = True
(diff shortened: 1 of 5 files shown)


dataclasses_json/mm.py
```diff
@@ -282,18 +282,19 @@ def schema(cls, mixin, infer_missing):
...
     #  https://marshmallow.readthedocs.io/en/stable/quickstart.html
     for field in dc_fields(cls):
-        metadata = (field.metadata or {}).get('dataclasses_json', {})
         metadata = overrides[field.name]
         if metadata.mm_field is not None:
             schema[field.name] = metadata.mm_field
         else:
             type_ = field.type
-            options = {}
+            options: typing.Dict[str, any] = {}
```
george-zubrienko, line 290: Should be `Any`?

george-zubrienko (commented):
Hi @deansg! I see that CI fails - could you take a look and try fixing it?

dataclasses_json/mm.py
deansg: Fixed

deansg:
@george-zubrienko Hi George, I fixed the build, but had to basically override the expected behavior in 2 tests for this to pass. I feel like the new behavior is a lot more correct, but technically this can be considered breaking.

george-zubrienko:
@lidatong will you have time to make a decision here? I'd like you to sign off either the merge as a regular patch, or as a breaking since test expectations are changed

deansg:
Basically my logic is: if you can't instantiate it as a Python object with an empty dictionary, then an empty JSON should also fail to be converted.
This is also the logic in the marshmallow_dataclass library, which today I use in addition to dataclass_json

george-zubrienko:
I believe this is a correct change, but I'll wait couple days more in case others have different opinions. If nothing comes in, I'll merge this as a bugfix around June 21

lidatong:
> @lidatong will you have time to make a decision here? I'd like you to sign off either the merge as a regular patch, or as a breaking since test expectations are changed

yes this makes sense to me as a fix

lidatong (dismissed):

deansg:
I see there's a conflict there that has an easy solution, do you want me to resolve it or is it better a maintainer do it since they are the ones that approve/merge? @george-zubrienko

george-zubrienko:
> I see there's a conflict there that has an easy solution, do you want me to resolve it or is it better a maintainer do it since they are the ones that approve/merge? @george-zubrienko 

@deansg Please go ahead and resolve conflicts for you PR, will merge right away

deansg:
@george-zubrienko resolved, seem like some styling check is failing now but it mentions a lot of errors unrelated to my PRs, maybe it's a new check?

george-zubrienko (approved):

george-zubrienko:
> @george-zubrienko resolved, seem like some styling check is failing now but it mentions a lot of errors unrelated to my PRs, maybe it's a new check?

Its all good, this check is optional and we'll make it green soon. You can help if you have time :)

PR #405, opened 2023-03-30
Use getter for Optional fields

ha1dfo:
Summary:
Use getter instead of [], so one can specify optional fields.

Test Plan:
Define a dataclass w/ an Optional field.
Feed it a Dict w/o that field present.
Observe exception.
Patch the code.
Do it again, observe the field is now a `None`

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -160,7 +160,7 @@ def _decode_dataclass(cls, kvs, infer_missing):
         if not field.init:
             continue
 
-        field_value = kvs[field.name]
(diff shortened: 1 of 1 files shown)


wQuole:
@ha1dfo May I ask why you closed this? :)

ha1dfo:
I don't quite remember; but ended up not using the lib in my project so I shelved the issue for a while.

wQuole:
Ah, I see. Did you find an alternative to dataclass-json, or did you not need the capabilities that dataclasses-json provides, anymore? :)

PR #409, opened 2023-06-01
feat: use default value in not optional field.

Bidaya0:
Case 1:
The upstream has passed a value that may be empty. This empty value has no business meaning. It is hoped that the dataclass_json layer can be completed instead of introducing more types to reduce the complexity of subsequent business logic.
```
{
  "a": null
}

{
  "a": ["1,2,3"]
}
```

```
a: Tuple[str] = ()
```


george-zubrienko:
Related to #410 

Sorry but I don't think this is a correct behaviour - see the reply in the issue

Issue #376, opened 2022-10-03
Leading underscore are not kept in `to_dict` method


clementta:
Hi,

Here is an example:
Having this sample:
```py
from typing import Optional
from dataclasses_json import LetterCase, Undefined, config, dataclass_json
from dataclasses import dataclass, field

@dataclass_json(undefined=Undefined.EXCLUDE, letter_case=LetterCase.CAMEL)
@dataclass
class Document:
    field_1: str
    field_2: Optional[str] = None
    _field_3: Optional[str] = None
    _field_4: Optional[int] = None
```

In version 0.5.4:
```py
Document(field_1="hello").to_dict()
#  {'field_1': 'hello', 'field_2': None, '_field_3': None, '_field_4': None}
```

In version 0.5.5 and above:
```py
Document(field_1="hello").to_dict()
#  {'field_1': 'hello', 'field_2': None, 'field_3': None, 'field_4': None}
```

I am expecting 0.5.4 and lower behavior.

* Is this a wanted feature ?
* Is there an config option to go back to the leading _ in dict ?

Thank you

matt035343:
I recreated this, and to me this is not a bug, because you are using camelCase. Underscores are not used in camelCase, and your example actually highlights a bug in the camelCase serialization; the number in the end should be spearated with an underscore.

I have opened a PR to fix this. Thanks!

PR #420, opened 2023-06-10
Fixes bug in string case serialization not taking numbers into account

matt035343:
As described in #376, the camelCase serialization does not remove underscores in front of numbers. I believe it should - let me know what you think.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -115,7 +115,10 @@ def _encode_overrides(kvs, overrides, encode_json=False):
             original_key = k
             k = letter_case(k) if letter_case is not None else k
             if k in override_kvs:
-                raise ValueError(f"Multiple fields map to the same JSON key after letter case encoding: {k}")
+                raise ValueError(
+                    f"Multiple fields map to the same JSON "
+                    f"key after letter case encoding: {k}"
+                )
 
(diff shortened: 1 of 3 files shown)


george-zubrienko (commented):
You can run into issues with this when having fields named a_1_1 vs a_11 - both will have the same name in Camel. I think underscores should be replaced by a-Z char instead

matt035343:
> You can run into issues with this when having fields named a_1_1 vs a_11 - both will have the same name in Camel. I think underscores should be replaced by any a-Z char instead OR you should check if camel will produce duplicate fields when applied, and runtime fail the serialisation.

imo, I think adding arbitrary chars is worse than underscores as it may also cause collisions with the original name. Then I think it is a better to raise an exception. Also, it is possible to provide custom field names in the metadata to manually avoid the conflicts.

george-zubrienko:
> a better to raise an exception

Agreed :)

george-zubrienko:
@matt035343 any chance you have time to add exception in case duplicate is produced and we merge? :)

matt035343:
> @matt035343 any chance you have time to add exception in case duplicate is produced and we merge? :)

Exception is now added, please take a look

george-zubrienko (approved):
Awesome!

PR #424, opened 2023-06-16
Add code_quality pipeline

s-vitaliy:


george-zubrienko (approved):
@lidatong fyi, we'll be fixing all mypy errors in the following commits and then this check should be made required :)

PR #426, opened 2023-06-16
fix: surface nested mapping errors

### LINKED ISSUE #425: [BUG] common errors from decoder on inner dataclass silently ignored
### Description

If you have a dataclass with a mapping field that includes a dataclass member, and you try to `from_dict` the outer dataclass, some errors from a custom `decoder` implementation will get silently ignored.

```py
from dataclasses_json import config, dataclass_json


def implementation_issue(value: str):
    raise TypeError("improper implementation")


@dataclass_json
@dataclass
class InnerExample:
    example_inner: int = field(metadata=config(decoder=implementation_issue))


@dataclass_json
@dataclass
class Example:
    example_field: dict[str, InnerExample]


# This should fail, but it doesn't. The new Example instance's example_field
# value is just empty instead.
Example.from_dict(dict(example_field=dict(test=dict(example_inner="what"))))
```

```
  File "/private/tmp/t.py", line 23, in <module>
    Example.from_dict(dict(example_field=dict(test=dict(example_inner="what"))))
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/api.py", line 70, in from_dict
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 210, in _decode_dataclass
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 279, in _decode_generic
    res = t(xs)
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 336, in <genexpr>
    items = (_decode_dataclass(type_arg, x, infer_missing)
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 196, in _decode_dataclass
    init_kwargs[field.name] = overrides[field.name].decoder(
  File "/private/tmp/t.py", line 6, in implementation_issue
    raise TypeError("improper implementation")
TypeError: improper implementation
```

3.9.16

### Environment description

irrelevant, cause is clear

skeggse:
Fixes #425.

### DIFF
--- a/tests/test_nested.py
+++ b/tests/test_nested.py
@@ -27,14 +27,8 @@ class TestDecoder:
                 DataClassXs([DataClassX(0), DataClassX(1)]))
 
     def test_nested_mapping_of_dataclasses(self):
-        err = None
-        try:
+        with pytest.raises(TypeError, match="positional arguments"):
             DataClassMappingBadDecode.from_dict(dict(map=dict(test=dict(id="irrelevant"))))
-        except TypeError as e:
-            if "positional arguments" not in e.args[0]:
-                raise
-            err = e
-        assert isinstance(err, TypeError)
 
 
 class TestNested:
--- a/tests/test_nested.py
+++ b/tests/test_nested.py
@@ -1,3 +1,4 @@
+import pytest
 from tests.entities import (DataClassMappingBadDecode,
                             DataClassWithDataClass,
                             DataClassWithList,
--- a/tests/test_nested.py
+++ b/tests/test_nested.py
@@ -1,4 +1,5 @@
-from tests.entities import (DataClassWithDataClass,
+from tests.entities import (DataClassMappingBadDecode,
+                            DataClassWithDataClass,
                             DataClassWithList,
                             DataClassWithNestedDictWithTupleKeys,
                             DataClassX,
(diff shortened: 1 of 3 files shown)


skeggse:
The mypy errors _look_ like they're already present on `master`, but I haven't carefully examined them.

george-zubrienko:
> The mypy errors _look_ like they're already present on `master`, but I haven't carefully examined them.

This is not a required check yet. You can ignore this for now :)

tests/test_nested.py
```diff
@@ -25,6 +26,16 @@ def test_nested_list_of_dataclasses(self):
         assert (DataClassXs.from_json('{"xs": [{"x": 0}, {"x": 1}]}') ==
                 DataClassXs([DataClassX(0), DataClassX(1)]))
 
+    def test_nested_mapping_of_dataclasses(self):
+        err = None
+        try:
+            DataClassMappingBadDecode.from_dict(dict(map=dict(test=dict(id="irrelevant"))))
```
s-vitaliy, line 32: I would recommend use `pytest.raises` for testing expected exceptions, here the docs: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

s-vitaliy (changes requested):
@skeggse, you can ignore errors in mypy for now, we will fix them. Let's fix the test and merge it :)

tests/test_nested.py
skeggse: i'm not terribly familiar with `pytest`, and i was hoping there would be something like that, thanks!

s-vitaliy (approved):

PR #435, opened 2023-07-13
Fixes heterogeneous tuple decoding

### LINKED ISSUE #427: [BUG] tuple decode assumes homogenous sequence
### Description

If I express a field type as `tuple[int, str]`, I would expect the two dimensions of the tuple to be decoded as each separate type. This is especially true if the type of the first dimension is an Enum, because assuming that only the first type applies to all values means decoding the second value as an Enum even though it isn't the appropriate value.

```py
from enum import Enum

from dataclasses_json import config, dataclass_json


class ExampleEnum(Enum):
    FIRST = "first"


@dataclass_json
@dataclass
class Example:
    example_inner: tuple[ExampleEnum, str]


# This should produce no error. On the contrary, this should be perfectly valid.
Example.from_dict(dict(example_inner=("first", "second")))
```

No error. Instead, got:
```py
  File "/tmp/t.py", line 18, in <module>
    Example.from_dict(dict(example_inner=("first", "second")))
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/api.py", line 70, in from_dict
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 210, in _decode_dataclass
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 279, in _decode_generic
    res = t(xs)
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 339, in <genexpr>
    items = (_decode_generic(type_arg, x, infer_missing) for x in xs)
  File "/usr/local/lib/python3.9/site-packages/dataclasses_json/core.py", line 258, in _decode_generic
    res = type_(value)
  File "/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/enum.py", line 384, in __call__
    return cls.__new__(cls, value)
  File "/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/enum.py", line 702, in __new__
    raise ve_exc
ValueError: 'second' is not a valid ExampleEnum
```

3.9.16

### Environment description

irrelevant, cause is obvious: lack of supported functionality, but in this case we should see an error due to no support for tuples rather than an error due to trying to decode the enum.

matt035343:
Fixes #427


george-zubrienko (commented):
Heterogenous nested tuples should also be tested I think:

```
Tuple[Tuple[int, str], Tuple[MyEnum, int], int]
```
for example

matt035343:
@george-zubrienko I have now added more tests

george-zubrienko (approved):
LGTM, pythoneers should be happy ;)

matt035343:
@rtsscy 

> hi there, does this break the definition of an infinite length of heterogenous list? ie: Tuple[str, ...]

Thanks for bringing this up! We did not consider this case at all. I have now created a new PR fixing this: https://github.com/lidatong/dataclasses-json/pull/440

george-zubrienko:
@rtsscy 
We have released a new version with a fix https://github.com/lidatong/dataclasses-json/releases/tag/v0.5.13

If you still see problems with this change, please open a new issue instead of commenting on the PR :)

Issue #398, opened 2022-12-25
Cannot generate schema for `typing.Tuple[...types]`


saabeilin:
Consider the following example:

```
import typing
from dataclasses import dataclass

from dataclasses_json import dataclass_json


@dataclass_json()
@dataclass
class Defaults:
    time_window: typing.Optional[typing.Tuple[int, int]] = None

Defaults.schema()
```

**Obtained result:**
```
  File "/home/......../.venv/lib/python3.10/site-packages/marshmallow/fields.py", line 809, in __init__
    super().__init__(*args, **kwargs)
TypeError: Field.__init__() takes 1 positional argument but 2 positional arguments (and 2 keyword-only arguments) were given
```

george-zubrienko:
I will retest this on latest and act accordingly

PR #440, opened 2023-07-19
Fixes bug related to Tuples defined with ellipsis

matt035343:
PR https://github.com/lidatong/dataclasses-json/pull/435 introduced a new bug that raises an exception when deserializing tuples defined with ellipsis (e.g. `Tuple[str, ...]`). This PR is fixing this issue.


george-zubrienko (approved):

deansg:
The bug still seems to happen, I have a `Tuple[str, ...]` as a field in a class and just calling `.schema()` on it fails with the following message:
`E       TypeError: Field.__init__() takes 1 positional argument but 2 positional arguments (and 3 keyword-only arguments) were given
`

deansg:
Full code to reproduce on Python 3.10 (very straightforward):

`
@dataclass_json
@dataclass
class SomeClass:
    tags: Tuple[str, ...] = None


def test_foo():
    SomeClass.schema()
`

matt035343:
Hi @deansg. If I am not mistaken, this seems to be related to an old issue (https://github.com/lidatong/dataclasses-json/issues/398) not directly related to what has been discussed in #440 and #435. I will have a look at #398 soon.

Issue #460, opened 2023-08-04
[BUG] Generated schema doesn't take into account custom decoder


deansg:
### Description

If a custom decoder is specified on a dataclass, the schema generated by the class' `schema()` method doesn't properly take it into account. This creates a situation where deserializing a dictionary through `from_dict` will work differently than through `schema().load(...)`.

### Code snippet that reproduces the issue

```
def split_str(data: str):
    return data.split(',')


@dataclass_json
@dataclass
class SomeClass:
    a: list[str] = field(
        default=None,
        metadata=config(
            decoder=split_str
        )
    )


def test_schemas():
    b = SomeClass.from_dict({'a': '1,2,3'})
    c = SomeClass.schema().load({'a': '1,2,3'})
```
In this example, variable b is deserialized correctly, while trying to deserialize using `Someclass.schema().load(...)` throws the following exception:

`marshmallow.exceptions.ValidationError: {'a': ['Not a valid list.']}`

### Describe the results you expected

Full consistency between .schema().load(...) and .from_dict(), both will take the custom decoder into account.

### Python version you are using

3.10.5

### Environment description

dataclasses-json==0.5.14

deansg:
Regarding the code quality check, that's a known mypy issue as seen here:
https://github.com/python/mypy/issues/2427

We can just add `--disable-error-code=method-assign` to the check if we want

PR #462, opened 2023-08-04
Fix: if field has custom decoder, schema takes it into account

deansg:
If a field has a custom decoder, replace its _deserialize logic with a dummy method, as it will be taken care of by the decoder.

Attempts to solve #460


deansg:
This is just an initial version. I'd like to know if I'm in the right direction or if this isn't it.

deansg:
Regarding the code quality check, that's a known mypy issue as seen here:
https://github.com/python/mypy/issues/2427

We can just add `--disable-error-code=method-assign` to the check if we want

dataclasses_json/mm.py
```diff
@@ -162,15 +163,16 @@ def __init__(self, *args, **kwargs):
             raise NotImplementedError()
 
         @typing.overload
-        def dump(self, obj: typing.List[A], many: typing.Optional[bool] = None) -> typing.List[TEncoded]:  # type: ignore
+        def dump(self, obj: typing.List[A],  # type: ignore
```
george-zubrienko, line 166: Please revert code formatting changes, so the PR contains functional changes only :)

dataclasses_json/mm.py
```diff
@@ -318,13 +320,26 @@ def schema(cls, mixin, infer_missing):
                 options['data_key'] = metadata.letter_case(field.name)
 
             t = build_type(type_, options, mixin, field, cls)
+            _account_for_decoder_if_necessary(field, t)
```
george-zubrienko, line 323:
```suggestion
            if field.metadata.get('dataclasses_json', {}).get('decoder'):
                t._deserialize = lambda v, *_a, **_kw: v
```

Also maybe add comment why a private field is being mutated here?

dataclasses_json/mm.py
```diff
@@ -318,13 +320,26 @@ def schema(cls, mixin, infer_missing):
...
+    decoder = field.metadata.get('dataclasses_json', {}).get('decoder')
+    if decoder:
+        # Used in order to avoid mypy error. See https://github.com/python/mypy/issues/2427 for more details
+        t._deserialize = _dummy_deserialize  # type: ignore
+
+
+def _dummy_deserialize(value: typing.Any, *_args, **_kwargs):
+    return value
+
+
```
george-zubrienko, line 342:
```suggestion
```

george-zubrienko (commented):
Thank you for contribution!

Minor comments, I don't think the approach is wrong, but I'm also not very string on marshmellow side, so a bit of explanation of this solution would be nice :)

deansg:
@george-zubrienko I reverted all formatting changes except for a single line which only contained spaces which IntelliJ just won't let me change at all, I hope it's not a deal breaker.
I also added an extended comment explaining the change.
Let me know if you have more comments.

george-zubrienko (approved):
Awesome, nice and clear! Thanks a lot for fixing this bug!

Issue #239, opened 2020-07-27
Union inside List or Dict is not deserialized as the corresponding type


a-mckinley:
Hi,

I think I have a similar issue to #82 and #222 in that Union types are deserialized to 'dict' as opposed to their expected type.  I am using version 0.5.1.

The marshmallow based workaround mentioned by @FaraSeer  in #82 does not appear to work as shown in my test case below.

Did I miss something here with the library usage?

Test case:

```python
from dataclasses import dataclass
from dataclasses_json import dataclass_json
from typing import Dict, Union, List
import json

@dataclass_json
@dataclass(frozen=True)
class TestChild:
    somefield: int = None

@dataclass_json
@dataclass(frozen=True)
class TestOtherChild:
    otherfield: int = None

@dataclass_json
@dataclass(frozen=True)
class NotWorkingDictUnion:
    d: Dict[str,Union[TestChild, TestOtherChild]]

@dataclass_json
@dataclass(frozen=True)
class NotWorkingListUnion:
    l: List[Union[TestChild, TestOtherChild]]

if __name__ == "__main__":
    data = {
        "d": {
            "child" : {
                "somefield" : 1
            },
            "other_child" : {
                "otherfield" : 2
            }
        }
    }
    json_str = NotWorkingDictUnion.schema().dump(data)
    obj = NotWorkingDictUnion.schema().load(json_str)

    # type is <class 'dict'>, as opposed to TestChild/TestOtherChild
    print(type(obj.d['child']))

    data = {
        "l": [
            {
                "somefield" : 1
            },
            {
                "otherfield" : 1
            }
        ]
    }
    json_str = NotWorkingListUnion.schema().dump(data)
    obj = NotWorkingListUnion.schema().load(json_str)

    # type is <class 'dict'>, as opposed to TestChild/TestOtherChild
    print(type(obj.l[0]))
```

geonnave:
I confirm this issue happening with version 0.5.2.

PR #464, opened 2023-08-07
Fix for #239: Union inside List or Dict is not deserialized as the correspond…

pawelwilczewski:
Fixed the issue described in #239.

I'm not sure this is the best way of fixing this as the comment suggests the type should have already been decoded but this is my best attempt.

Frankly, I haven't tested it properly (works for my use cases) but it seems to me the logic is valid. 

I am not very familiar with the repo, so please let me know if I'm violating some of the rules set here and I'll try to adhere.


pawelwilczewski:
It would be good to add some test scenarios like the one in the mentioned issue.

george-zubrienko (commented):
Hello @pawelwilczewski ! Thank you for the contribution - please add unit tests that back your claim the issue is fixed - and re-request a review ;)

pawelwilczewski:
Stole a little bit of test code from the issue - hopefully the author a-mckinley doesn't mind

pawelwilczewski:
@george-zubrienko done! Let me know if anything else needs changing!

pawelwilczewski:
I noticed it won't work with `.schema.loads()` because I never made changes to the `mm.py`. I'm afraid I won't have free time to get around to doing that in the nearest future...

pawelwilczewski:
I'd think a fix similar to what I did could be implemented in the `else` clause in line 87 in mm.py

george-zubrienko:
> I'd think a fix similar to what I did could be implemented in the `else` clause in line 87 in mm.py

I'll review what is here this week, then we'll open a new issue and somebody can pick it up ;)

TristanSpeakEasy:
> > I'd think a fix similar to what I did could be implemented in the `else` clause in line 87 in mm.py
> 
> I'll review what is here this week, then we'll open a new issue and somebody can pick it up ;)

would love to see this one go through as we have another PR that adds more support for undiscriminated unions we would like to go in after this https://github.com/pawelwilczewski/dataclasses-json/pull/1

pawelwilczewski:
Not too sure why the poetry installation might have failed...

george-zubrienko:
Merging https://github.com/lidatong/dataclasses-json/pull/476 will fix, sorry for delays in reviewing, on my todo

pawelwilczewski:
No worries at all, and no pressure!

dataclasses_json/core.py
```diff
@@ -310,8 +310,27 @@ def _decode_generic(type_, value, infer_missing):
...
+                for type_option in type_options:
+                    if is_dataclass(type_option):
+                        res = _decode_dataclass(type_option, value, infer_missing)
+                        changed = True
+                        break
+                if not changed:
+                    warnings.warn(
+                        f"Failed encoding {value} Union dataclasses."
+                        f"Expected Union to include a dataclass and it didn't."
+                    )
```
george-zubrienko, line 333:
```suggestion
            type_options = _get_type_args(type_)
            res = value
            if type(value) not dict or dict not in type_options:
                # FIXME if all types in the union are dataclasses this
                #  will just pick the first option -
                #  maybe find the best fitting class in that case instead?
                for type_option in type_options:
                    if is_dataclass(type_option):
                        res = _decode_dataclass(type_option, value, infer_missing)
                        break
                if res == value:
                    warnings.warn(
                        f"Failed to encode value {value} of type {type_}."
                        f"Expected Union to include a dataclass and it didn't."
                    )
```

george-zubrienko (dismissed):
LGTM, minor refactoring suggestion

pawelwilczewski:
Seems like there were some issues with the refactor. I'll roll back and you can automatically merge - unless you'd like to give it another go. Cheers!

george-zubrienko:
I'll take a look why and suggest a new one. The one I added was from the phone so might have missed the `if` condition

pawelwilczewski:
Yeah I gave it an attempt but kinda just skimmed through it relying on github actions rather than properly testing

george-zubrienko:
If you can fix would be great, if not I'll prepare a proper suggestion within next 2 days. I want this to be included in 0.6.0 :)

pawelwilczewski:
I believe this does it now

george-zubrienko (approved):

george-zubrienko:
Awesome, thanks a lot for the contribution, will be released this weekend!

pawelwilczewski:
Happy to contribute. Thanks a lot for your overall and this PR's input!

PR #468, opened 2023-08-11
Narrow down implicit boolean conversion introduced in 0.5.8

### LINKED ISSUE #466: [BUG] Type coercion since 0.5.9
### Description

I'll let you guys explain that this is intended and why, but starting with 0.5.9 there is type coercion going on when using the `from_dict()` method.

A non-zero int becomes True on a bool field, for instance. An int becomes its stringified version on a string field.

This feels to me like a breaking change, but I don't see this anywhere in the release notes. Nor in the version number in any case.

Was this intended? If so, why? 

```python

@dataclass
class MyDataclass(DataClassJsonMixin):
    this_must_be_bool: bool
    this_must_be_string: str


d = {"this_must_be_bool": 1234, "this_must_be_string": 1234}
c = MyDataclass.from_dict(d)

assert isinstance(c.this_must_be_bool, bool)
assert isinstance(c.this_must_be_string, str)
```

So this snippet raises with 0.5.8. Anything after and including 0.5.9 does _not_ raise.

3.11.4

### Environment description

Tested with dataclasses-json 0.5.9, 0.5.12 and 0.5.14

george-zubrienko:
Fixes #466 (at least the boolean part of it that is defo a bug)

Enhanced builtin instantiation introduced in 0.5.8 had a side effect of converting any int in json to a `True` boolean value if the destination field has `bool` type. This PR narrows allowed value range for implicit conversion to boolean


tests/test_builtins.py
```diff
@@ -13,22 +14,53 @@ class DataClassWithBuiltins(DataClassJsonMixin):
...
+def test_dataclass_with_implicit_builtins(model_dict: Dict, expected_model: DataClassWithBuiltins):
     assert DataClassWithBuiltins.from_dict(model_dict) == expected_model
+
+
+@mark.parametrize(
+    "model_dict",
+    [
+        param(
+            {"actually_a_str": "str", "actually_an_int": 42, "actually_a_float": 42.1, "actually_a_bool": 1234},
+            id="Bool passed using valid discrete integer range [0,1]"
```
george-zubrienko, line 56:
```suggestion
            id="Bool passed using invalid integer"
```

dataclasses_json/core.py
```diff
@@ -244,7 +244,16 @@ def _support_extended_types(field_type, field_value):
...
                if isinstance(field_value, UUID)
                else UUID(field_value))
-    elif _issubclass_safe(field_type, (int, float, str, bool)):
+    elif _issubclass_safe(field_type, bool):
+        # issubclass(bool, int) -> True
+        # thus values >1 will always convert to True in the next clause, unless intercepted
+        if field_value in ('True', 'true', True, 1):
+            res = True
+        elif field_value in ('False', 'false', False, 0):
+            res = False
```
matt035343, line 253: In JavaScript (where JSON is born out of), empty list, empty dict and empty string are all considered false, and if they just ahve one or more entry, they are true.

dataclasses_json/core.py
```diff
@@ -244,7 +244,16 @@ def _support_extended_types(field_type, field_value):
         res = (field_value
                if isinstance(field_value, UUID)
                else UUID(field_value))
-    elif _issubclass_safe(field_type, (int, float, str, bool)):
+    elif _issubclass_safe(field_type, bool):
+        # issubclass(bool, int) -> True
+        # thus values >1 will always convert to True in the next clause, unless intercepted
```
matt035343, line 249: I do not see >1 ints being set to true, or am I blind?

dataclasses_json/core.py
```diff
@@ -244,7 +244,16 @@ def _support_extended_types(field_type, field_value):
...
+        if field_value in ('True', 'true', True, 1):
+            res = True
+        elif field_value in ('False', 'false', False, 0):
+            res = False
+        else:
+            raise ValueError(f"Value {field_value} of input type {field_type.__name__} cannot be decoded as boolean")
+    elif _issubclass_safe(field_type, int) or _issubclass_safe(field_type, float) or _issubclass_safe(field_type, str):
```
matt035343, line 256: Please keep  the tuple notation here

george-zubrienko: This is not correct because that method doesn't accept tuple. It was an error in original implementation

dataclasses_json/core.py
george-zubrienko: Run code from the issue author with 0.5.14 and you will set the 1234 int field being set to True when running `from_dict`

dataclasses_json/core.py
george-zubrienko: Empty lists/dicts/strings are normally not decoded as booleans, even though in JS that is a common place, where JSON is used, this behaviour is avoided.

dataclasses_json/core.py
matt035343: The argument is passed directly to `issubclass` which accepts a tuple.
https://python-reference.readthedocs.io/en/latest/docs/functions/issubclass.html

dataclasses_json/core.py
matt035343: How is that caught by your if-statements below?

dataclasses_json/core.py
matt035343: As I do not like type juggling myself, I will rest this case

dataclasses_json/core.py
george-zubrienko: It is not caught, we just do not allow the invocation to happen in the first place, if target type is bool

dataclasses_json/core.py
george-zubrienko: I see, however, there is one twist there: https://docs.python.org/3.11/library/functions.html#issubclass

> Changed in version 3.10: classinfo can be a [Union Type](https://docs.python.org/3.11/library/stdtypes.html#types-union).

I'll see if this affects anything

dataclasses_json/core.py
matt035343: Also, I was mistaken,  empty list and empty dict are coerced to true in JS

george-zubrienko:
Closing as this is too risky to integrate, will instead do a separate one for v1

PR #469, opened 2023-08-12
Fixing case where multiple fields have the same field name

### LINKED ISSUE #270: Use the same field_name in more than one field
I have a use case in which a field in a JSON structure should be mapped into two separate fields in my dataclass:

```json
{
    "myJsonField": {
        "first": "val1",
        "second": "val2"
    },
    "otherField": "valotherfield"
}
```

I have tried using specialized decoders for each field, so I can extract the portion of what I need. In this case, I want to avoid creating another dataclass to decode the field "myJsonField":

```python

contents = "{...}"  # the JSON data above

def first_decoder(data: Dict[str, str]) -> str:
    return data["first"]

def second_decoder(data: Dict[str, str]) -> str:
    return data["second"]

@dataclass_json
@dataclass
class MyDataClass:
    first: str = field(metadata=config(field_name="myJsonField", encoder=first_decoder))
    second: str = field(metadata=config(field_name="myJsonField", encoder=second_decoder))
    other_field: str = field(metadata=config(field_name="other_Field"))

MyDataClass.from_json(contents)
```

The above does not work, as the second field overrides the first field.

What is the best way to tackle this case? Is there an alternative other than creating nested dataclasses?

matt035343:
Fixes #270

Fixing case where multiple fields have the same field name by inverting the map holding the field name overrides

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -1,11 +1,12 @@
 import datetime
+import json
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Optional, Dict
 
 import pytest
 from marshmallow import fields, ValidationError
 
-from dataclasses_json import DataClassJsonMixin
+from dataclasses_json import DataClassJsonMixin, dataclass_json, config
 
 
 @dataclass
@@ -24,7 +25,8 @@ class StringDate(DataClassJsonMixin):
             'decoder': str,
             'mm_field': fields.String(required=False)}
         })
-    
+
+
 @dataclass
 class OptionalStringDate(DataClassJsonMixin):
     string_date: Optional[datetime.datetime] = field(
@@ -34,6 +36,14 @@ class OptionalStringDate(DataClassJsonMixin):
             'decoder': str,
             'mm_field': fields.String(required=False)}
         })
+
+
+@dataclass_json
+@dataclass
+class MyDataClass:
+    first: str = field(metadata=config(field_name="myJsonField", decoder=lambda x: x["first"]))
+    second: str = field(metadata=config(field_name="myJsonField", decoder=lambda x: x["second"]))
+    other_field: str = field(metadata=config(field_name="otherField"))
 
 
 car_schema = Car.schema()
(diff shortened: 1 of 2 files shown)


tests/test_metadata.py
```diff
@@ -36,6 +38,14 @@ class OptionalStringDate(DataClassJsonMixin):
         })
 
 
+@dataclass_json
+@dataclass
+class MyDataClass:
+    first: str = field(metadata=config(field_name="myJsonField", decoder=lambda x: x["first"]))
+    second: str = field(metadata=config(field_name="myJsonField", decoder=lambda x: x["second"]))
+    other_field: str = field(metadata=config(field_name="otherField"))
```
george-zubrienko, line 46: I'm not sure, but is this a good test case? Like, declaring `myJsonField` as a regular dict will not trigger the fixed behaviour first place?

matt035343: @george-zubrienko  I am not sure I understand. Are you suggesting to use a dataclass instead of a dict?

george-zubrienko: I mean the `MyDataClass` can be defined like this:

```
class MyDataClass:
    myJsonField: dict
    other_field: str = field(metadata=config(field_name="otherField"))
    
    @property
    def first(self) -> str:
        return self.myJsonField.get('first', '')

    @property
    def second(self) -> str:
        return self.myJsonField.get('first', '')
```

george-zubrienko: Just IMO, the way test case is defined, dataclass does not reflect the actual data structure being read, but rather relies on some rather artificial deserializing behaviour, which for example in v1 warrants a custom serializer for the class. If this is something people **actually** do in their user code, I'd like to see examples. I am near 100% certain we will not support these kind of cases in v1 and force people to write custom serializer, thus I'd be wary of using these as test cases for v0 as well.

matt035343: @george-zubrienko This is PR spawned from https://github.com/lidatong/dataclasses-json/issues/270

matt035343: @george-zubrienko and I discussed that it is actually a bad practice do like #270 does. It will only work one way (decoding only, not encoding). Instead, use the approach @george-zubrienko outlined above.

I will discontinue this PR and make a new one that throws an exception instead.

PR #470, opened 2023-08-12
Add documentation for how to use with Numpy and Pandas

### LINKED ISSUE #90: Add ability to encode numpy datatypes
Hi,

in my program most of calculations are done with numpy and thus values are e.g of type np.int32 or numpy.ndarrays. 

Consider following class:
```
@dataclass_json
@dataclass
class Results:
     value: int
     values: List[int]

results = Results()
results.value = some_numpy_computation_returning_numpy_int()
results.values = some_numpy_computation_returning_numpy_array()
json_data = results.to_json()
```
When running this code I would get a conversion error.

It would be very convenient if dataclassjson would be able to handle those errors, by passing in an extended json encoder which handles numpy datatypes.

In the dataclassjson code some cases are already dealt with the _ExtendedEncoder.
```
class _ExtendedEncoder(json.JSONEncoder):
    def default(self, o) -> JSON:
        result: JSON
        if _isinstance_safe(o, Collection):
            if _isinstance_safe(o, Mapping):
                result = dict(o)
            else:
                result = list(o)
        elif _isinstance_safe(o, datetime):
            result = o.timestamp()
        elif _isinstance_safe(o, UUID):
            result = str(o)
        elif _isinstance_safe(o, Enum):
            result = o.value
        else:
            result = json.JSONEncoder.default(self, o)
        return result
```
However, I cannot extend this decoder to support my needs, since cls=_ExtendedEncoder is not in the argument list of to_json.
 
```
 class DataClassJsonMixin(abc.ABC):
    """
    DataClassJsonMixin is an ABC that functions as a Mixin.

    As with other ABCs, it should not be instantiated directly.
    """

    def to_json(self,
                *,
                skipkeys: bool = False,
                ensure_ascii: bool = True,
                check_circular: bool = True,
                allow_nan: bool = True,
                indent: Optional[Union[int, str]] = None,
                separators: Tuple[str, str] = None,
                default: Callable = None,
                sort_keys: bool = False,
                **kw) -> str:
        return json.dumps(_asdict(self),
                          cls=_ExtendedEncoder,
                          skipkeys=skipkeys,
                          ensure_ascii=ensure_ascii,
                          check_circular=check_circular,
                          allow_nan=allow_nan,
                          indent=indent,
                          separators=separators,
                          default=default,
                          sort_keys=sort_keys,
                          **kw)
```
Would it be possible to change it to:
```
 class DataClassJsonMixin(abc.ABC):
    """
    DataClassJsonMixin is an ABC that functions as a Mixin.

    As with other ABCs, it should not be instantiated directly.
    """

    def to_json(self,
                *,
                skipkeys: bool = False,
                ensure_ascii: bool = True,
                check_circular: bool = True,
                allow_nan: bool = True,
                indent: Optional[Union[int, str]] = None,
                separators: Tuple[str, str] = None,
                default: Callable = None,
                sort_keys: bool = False,
                cls: json.JSONEncoder=_ExtendedEncoder,
                **kw) -> str:
        return json.dumps(_asdict(self),
                          cls=cls,
                          skipkeys=skipkeys,
                          ensure_ascii=ensure_ascii,
                          check_circular=check_circular,
                          allow_nan=allow_nan,
                          indent=indent,
                          separators=separators,
                          default=default,
                          sort_keys=sort_keys,
                          **kw)
```
I know about the possibility to define an encoder via metadata, but instead of adding this to all of my fields of the dataclasses, I'd rather prefer a global solution.

matt035343:
Closes #90 and #352 

I apologize for the commit message. I forgot to change it before comitting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,8 +434,7 @@ from __future__ import annotations
 as it will cause problems with the way dataclasses_json accesses the type annotations.
 
 ### Use numpy or pandas types?
-Numpy and pandas types are not supported by default, but you can easily add support for them 
-by using custom decoders and encoders. Below are two examples for numpy and pandas types.
+Data types specific to libraries commonly used in data analysis and machine learning like [numpy](https://github.com/numpy/numpy) and [pandas](https://github.com/pandas-dev/pandas) are not supported by default, but you can easily enable them by using custom decoders and encoders. Below are two examples for `numpy` and `pandas` types.
 
 ```python
 from dataclasses import field, dataclass
--- a/README.md
+++ b/README.md
@@ -433,6 +433,36 @@ from __future__ import annotations
 ```
 as it will cause problems with the way dataclasses_json accesses the type annotations.
 
+### Use numpy or pandas types?
+Numpy and pandas types are not supported by default, but you can easily add support for them 
+by using custom decoders and encoders. Below are two examples for numpy and pandas types.
+
+```python
+from dataclasses import field, dataclass
+from dataclasses_json import config, dataclass_json
+import numpy as np
+import pandas as pd
+
+@dataclass_json
+@dataclass
+class DataWithNumpy:
+    my_int: np.int64 = field(metadata=config(decoder=np.int64))
+    my_float: np.float64 = field(metadata=config(decoder=np.float64))
+    my_array: np.ndarray = field(metadata=config(decoder=np.asarray))
+DataWithNumpy.from_json("{\"my_int\": 42, \"my_float\": 13.37, \"my_array\": [1,2,3]}")
+
+@dataclass_json
+@dataclass
+class DataWithPandas:
+    my_df: pd.DataFrame = field(metadata=config(decoder=pd.DataFrame.from_records, encoder=lambda x: x.to_dict(orient="records")))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -433,6 +433,36 @@ from __future__ import annotations
 ```
 as it will cause problems with the way dataclasses_json accesses the type annotations.
 
+### Use numpy or pandas types?
+Numpy and pandas types are not supported by default, but you can easily add support for them 
+by using custom decoders and encoders. Below are two examples for numpy and pandas types.
```
george-zubrienko, line 438:
```suggestion
Data types specific to libraries commonly used in data analysis and machine learning like [numpy](https://github.com/numpy/numpy) and [pandas](https://github.com/pandas-dev/pandas) are supported by default, but you can easily enable them by using custom decoders and encoders. Below are two examples for `numpy` and `pandas` types.
```

george-zubrienko (dismissed):

matt035343:
@george-zubrienko Please approve again, reviews are reset upon commit apparently

george-zubrienko (approved):

PR #471, opened 2023-08-12
Support for Counter objects

matt035343:
Based on https://github.com/lidatong/dataclasses-json/pull/363, I have resolved the conflicts and fixed a bug I found when updating the rest of the code base.

@n1chre please have a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,35 @@ from __future__ import annotations
 ```
 as it will cause problems with the way dataclasses_json accesses the type annotations.
 
+### Use numpy or pandas types?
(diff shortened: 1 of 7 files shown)


george-zubrienko (commented):
https://docs.python.org/3/library/collections.html#collections.Counter

afaik typing.Counter is deprecated and Counter is just another python sugar for a dictionary type. Does this affect implementation, if taken into account?

george-zubrienko:
> @george-zubrienko Since it is only the typing alias that is deprecated I can use collections.Counter instead. I still think support for Counter is relevant even though it is just a wrapper.

Yeah I do not deny this is needed, I'm just curious if we can fallback to code that handles dict instead of writing logic specific to `Counter` type, given this might break due to import being removed in future python versions?

george-zubrienko (approved):
Bless `typing` lib :D

matt035343:
I will merge this by the end of this week unless @n1chre raises any concerns

Issue #442, opened 2023-07-20
[FEATURE] Graduate to v1.0 API


george-zubrienko:
### Description

Based on some discussions we see in issues around like #433 , #243, #139, #106 and probably more, I can pinpoint a few common pains:

1. Users are generally confused about whether they should use annotation `@dataclass_json` or the subclass approach with `DataClassJsonMixin`, the latter being similar to what `mashumaro` does. There are also a few things that can only be done from the annotation or global/field level config, most common example being `letter_case=...`.
2. Annotation usage leads to problems with static code analysis and thus IDE autocompletion tools - from personal experience, `pylint` hates it so in our projects we always use **both**, which adds boilerplate and confusion, especially for new developers. Now we can also see from both recent and past issues that people in general have to be creative to make their static code analysis not hate the lib. This is a usability problem and I think it must be addressed one way or another.
3. Mutating user classes via annotation or forcing people to subclass may lead to name conflicts or incorrect field resolution for cases like #106 (note this is usually not a problem in any non-python json ser-deser lib for python reasons)

I'd like to propose a breaking API change in 1.0 release to get rid of those issues, and potentially greatly improve user experience.

### Possible solution

Consider current quickstart:

```
from dataclasses import dataclass
from dataclasses_json import dataclass_json


@dataclass_json
@dataclass
class Person:
    name: str


person = Person(name='lidatong')
person.to_json()  # '{"name": "lidatong"}' <- this is a string
person.to_dict()  # {'name': 'lidatong'} <- this is a dict
Person.from_json('{"name": "lidatong"}')  # Person(1)
Person.from_dict({'name': 'lidatong'})  # Person(1)

# You can also apply _schema validation_ using an alternative API
# This can be useful for "typed" Python code

Person.from_json('{"name": 42}')  # This is ok. 42 is not a `str`, but
                                  # dataclass creation does not validate types
```
Change it to this:

```
from dataclasses import dataclass
from dataclasses_json.functions import from_json, to_json, to_dict, from_dict

@dataclass
class DataPerson:
    name: str

class RegularPerson:
    name: str

d_person = DataPerson(name='lidatong')
r_person = RegularPerson(name='foobar')

# '{"name": "lidatong"}' <- this is a string
# internally calls json.dumps(to_dict(d_person))
to_json(d_person )  

# {'name': 'lidatong'} <- this is a dict
to_dict(d_person)

# TypeError: r_person is not a dataclass
to_json(r_person )
to_dict(r_person )


[...]

PR #475, opened 2023-08-19
Improve dataclass_json and _process_class type annotations

ringohoffman:
Use `typing.overload` to differentiate the return type of `dataclass_json` based on whether `_cls` parameter is `None` or not.

[This is motivated by fixing type issues in flytekit, which does not only use `dataclass_json` as a decorator](https://github.com/flyteorg/flytekit/blob/6acc97b91851b2e9c139d73ed763a085a53b32b0/flytekit/core/type_engine.py#L1550).

```python

from dataclasses_json.api import dataclass_json


@dataclass
class Point:
    x: int
    y: int


directly_wrapped = dataclass_json(Point)

reveal_type(directly_wrapped)
# Before:
# pyright: Type of "directly_wrapped" is "((cls: Any) -> Any) | Any"
# mypy: Revealed type is "Any"
# PyCharm: directly_wrapped: (cls: {to_json, from_json, to_dict, from_dict, schema, __init__, dataclass_json_config}) -> {to_json, from_json, to_dict, from_dict, schema, __init__, dataclass_json_config} | {to_json, from_json, to_dict, from_dict, schema, __init__, dataclass_json_config} = dataclass_json(Point)

# After:
# pyright: Type of "dataclass_json(Point)" is "type[Point]"
# mypy: Revealed type is "Type[a.Point]"
# PyCharm: directly_wrapped: Type[Point] = dataclass_json(Point)

indirectly_wrapped = dataclass_json()(Point)

reveal_type(indirectly_wrapped)
# Before:
# pyright: Type of "indirectly_wrapped" is "Any"
# mypy: Revealed type is "Any"
# PyCharm: indirectly_wrapped: {to_json, from_json, to_dict, from_dict, schema, __init__, dataclass_json_config} = dataclass_json()(Point)

# After:
# pyright: Type of "dataclass_json()(Point)" is "type[Point]"
# mypy: Revealed type is "Type[a.Point]"
# PyCharm: indirectly_wrapped: Type[Point] = dataclass_json()(Point)

####################################################################################################


@dataclass_json
@dataclass
class Point2:
    x: int
    y: int


reveal_type(Point2)
# Before:
# pyright: Type of "Point2" is "type[Point2]"
# mypy: Revealed type is "def (x: builtins.int, y: builtins.int) -> a.Point2"
# PyCharm: class Point2

# After:
# pyright: Type of "Point2" is "type[Point2]"
# mypy: Revealed type is "def (x: builtins.int, y: builtins.int) -> a.Point2"
# PyCharm: class Point2


####################################################################################################


@dataclass_json()
@dataclass
class Point3:
    x: int
    y: int

reveal_type(Point3)
# Before:
# pyright: Type of "Point3" is "type[Point3]"
# mypy: Revealed type is "def (x: builtins.int, y: builtins.int) -> a.Point3"
# PyCharm: class Point3

# After:
# pyright: Type of "Point3" is "type[Point3]"
# mypy: Revealed type is "def (x: builtins.int, y: builtins.int) -> a.Point3"
# PyCharm: class Point3

```

### DIFF
--- a/dataclasses_json/api.py
+++ b/dataclasses_json/api.py
@@ -1,9 +1,8 @@
 import abc
 import json
-from typing import (Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar,
-                    Union)
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar, Union, overload
 
-from dataclasses_json.cfg import config, LetterCase  # noqa: F401
+from dataclasses_json.cfg import config, LetterCase
 from dataclasses_json.core import (Json, _ExtendedEncoder, _asdict,
                                    _decode_dataclass)
 from dataclasses_json.mm import (JsonData, SchemaType, build_schema)
@@ -12,6 +11,7 @@ from dataclasses_json.utils import (_handle_undefined_parameters_safe,
                                     _undefined_parameter_action_safe)
 
 A = TypeVar('A', bound="DataClassJsonMixin")
+T = TypeVar('T')
 Fields = List[Tuple[str, Any]]
 
 
@@ -102,8 +102,18 @@ class DataClassJsonMixin(abc.ABC):
                       unknown=unknown)
 
 
-def dataclass_json(_cls=None, *, letter_case=None,
-                   undefined: Optional[Union[str, Undefined]] = None):
+@overload
+def dataclass_json(_cls: None = ..., *, letter_case: Optional[LetterCase] = ...,
+                   undefined: Optional[Union[str, Undefined]] = ...) -> Callable[[Type[T]], Type[T]]: ...
+
+
+@overload
+def dataclass_json(_cls: Type[T], *, letter_case: Optional[LetterCase] = ...,
+                   undefined: Optional[Union[str, Undefined]] = ...) -> Type[T]: ...
+
+
+def dataclass_json(_cls: Optional[Type[T]] = None, *, letter_case: Optional[LetterCase] = None,
+                   undefined: Optional[Union[str, Undefined]] = None) -> Union[Callable[[Type[T]], Type[T]], Type[T]]:
     """
     Based on the code in the `dataclasses` module to handle optional-parens
     decorators. See example below:
@@ -114,7 +124,7 @@ def dataclass_json(_cls=None, *, letter_case=None,
         ...
     """
 
-    def wrap(cls):
+    def wrap(cls: Type[T]) -> Type[T]:
         return _process_class(cls, letter_case, undefined)
 
     if _cls is None:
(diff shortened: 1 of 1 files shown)


george-zubrienko:
Hi @ringohoffman! If you can provide additional information on how this change interacts with:

- Pylance (VS code helper)
- PyCharm editor
- PyLint (adding a simple test step that runs pylint would be fine)
- Pyright (adding a simple test step that runs pylint would fine)

This would significantly speed up the review process here, otherwise I'll have to try all these myself before I can review. Community had issues with this codepiece some time ago, so I have to be extra careful here. Also a sidenote, `@overload` functionality is not a very stable thing to rely on for such change, but let's put that aside and get basic checks done first.

ringohoffman:
What I had already posted were the changes you can see in VS Code using pylance/pyright. Pylance is a Visual Studio Code extension that uses pyright at its core for static type checking. I've gone ahead and added the before and after for PyCharm and mypy to the description as well.

> adding a simple test step that runs pylint/pyright would be fine

What do you mean by this? Do you want me to add a GitHub Workflow step to run pyright and pylint?

When I run pyright on the `dataclasses_json` module, there are 25 pre-existing errors. This PR doesn't add to or resolve these errors. I don't think that adding a workflow step for this is feasible at this time. Moreover, I think the poor type-checking compatibility of the `dataclass_json` decorator would completely preclude us from enforcing no pyright errors on this codebase.

I don't see any [`typecheck`](https://docs.pylint.org/features.html#typecheck-checker) warnings/errors when I run pylint on the code in my description, before or after. This PR does resolve a pylint warning for the unused `LetterCase` import in dataclasses_json/api.py since we are using it in the type hints now. Given that pylint didn't catch anything related to these changes, I see adding pylint to the workflow for this repository as a separate issue, but one that I would support.

Let me know how you want to move forward based on this and my updated description.

george-zubrienko:
I think it's ok, however I'm curious if this works with IDE autocomplete? You snippets are runtime code, but autocomplete does not rely on reveal_type etc. Though people **mainly** complain about static type checkers, we had cases when they also want IDE autocomplete to work like it does when using a Mixin.

Other small question, we plan to deprecate this function in v1 API (coming soon). How does that affect you?

ringohoffman:
> I think it's ok, however I'm curious if this works with IDE autocomplete? You snippets are runtime code, but autocomplete does not rely on reveal_type etc. Though people **mainly** complain about static type checkers, we had cases when they also want IDE autocomplete to work like it does when using a Mixin.
> 
> Other small question, we plan to deprecate this function in v1 API (coming soon). How does that affect you?

I don't think it's possible to make `dataclass_json` compatible with auto-complete. I think you have to choose the return type as either the wrapped type or as `DataClassJsonMixin`, but either way you will lose access to part of the actually available runtime methods. [I asked the author of pyright how he would handle making this function static-typing friendly](https://github.com/microsoft/pyright/discussions/5763), and he had recommended to just use the mixin as a mixin instead of the decorator.

This doesn't impact me in any major way. I already migrated `flytekit` to use `DataClassJsonMixin` as a mixin instead of  the `dataclass_json` decorator: https://github.com/flyteorg/flytekit/pull/1801 They only use `dataclass_json` in that one place now: https://github.com/flyteorg/flytekit/blob/0a772f3621ce9bd1a3e3b2f1e691458f25a31876/flytekit/core/type_engine.py#L1549

george-zubrienko (approved):

george-zubrienko:
> This doesn't impact me in any major way. I already migrated flytekit to use DataClassJsonMixin

Good, though v1 deprecates both the Mixin and the annotation - check discussion in #442. Note that we'll release v1 alongside v0 and deprecation of v0 will take **very long** time so please do not be scared :) But it helps a lot for us to understand the use cases like the code snippet you referenced.

Thanks a lot for contribution, we have a bit of an issue with 3.7 rn, but I plan to go around it and release your commit in 0.6 in a few days.

PR #478, opened 2023-09-04
improve Union deserialization when "__type" field specifier is not present

idbentley:
This pull request adds a best-effort deserialization to Union types when no `__type` field specifier is present.

In the case that no `__type` field is present, a warning is issued describing the risk of such unmarshalling attempt.

We attempt to decode ensuring that keys exist on the type, and the values are appropriate.  If either of these conditions fails, we abandon that dataclass type, and move onto the next one.

If no type is found, a warning is issued, the value is not decoded, and is left as-is.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -312,18 +312,23 @@ def _decode_generic(type_, value, infer_missing):
                 res = _support_extended_types(type_arg, value)
         else:  # Union (already decoded or try to decode a dataclass)
             type_options = _get_type_args(type_)
-            res = value  # assume already decoded
-            if type(value) is dict and dict not in type_options:
-                # FIXME if all types in the union are dataclasses this
-                #  will just pick the first option -
-                #  maybe find the best fitting class in that case instead?
+            if type(value) is not dict or dict in type_options:
+                # already decoded
+                res = value
+            else:
+                changed = False
                 for type_option in type_options:
                     if is_dataclass(type_option):
-                        res = _decode_dataclass(type_option, value, infer_missing)
-                        break
-                if res == value:
+                        try:
+                            res = _decode_dataclass(type_option, value, infer_missing)
+                            changed = True
+                            break
+                        except (KeyError, ValueError):
+                            continue
+                if not changed:
+                    res = value
                     warnings.warn(
-                        f"Failed to encode {value} Union dataclasses."
+                        f"Failed encoding {value} Union dataclasses."
                         f"Expected Union to include a dataclass and it didn't."
                     )
     return res
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -314,19 +314,17 @@ def _decode_generic(type_, value, infer_missing):
             type_options = _get_type_args(type_)
             res = value  # assume already decoded
             if type(value) is dict and dict not in type_options:
-                changed = False
                 for type_option in type_options:
                     if is_dataclass(type_option):
                         try:
                             res = _decode_dataclass(type_option, value, infer_missing)
-                            changed = True
                             break
                         except (KeyError, ValueError):
                             continue
-                if not changed:
+                if res == value:
                     warnings.warn(
-                        f"Failed to encode {value} Union dataclasses."
-                        f"Expected Union to include a dataclass and it didn't."
+                        f"Failed to decode {value} Union dataclasses."
+                        f"Expected Union to include a matching dataclass and it didn't."
                     )
     return res
 
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -325,7 +325,7 @@ def _decode_generic(type_, value, infer_missing):
                             continue
                 if not changed:
                     warnings.warn(
-                        f"Failed encoding {value} Union dataclasses."
+                        f"Failed to encode {value} Union dataclasses."
                         f"Expected Union to include a dataclass and it didn't."
                     )
     return res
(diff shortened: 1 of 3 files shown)


idbentley:
@george-zubrienko and @pawelwilczewski Building on #464 I added simple support for iterating over available `dataclasses` when deserializing objects withou `__type` specifiers.

george-zubrienko (commented):
Hello, thank you for contribution and nice to see people are actively collaborating! From what I see `core.py` it seems like your fork was behind master? It looks like the original commit from @pawelwilczewski ?

idbentley:
I don't see what you mean @george-zubrienko .  The logic change I added in `core.py` is adding an `if` check for each dataclass in the `type_options`, which catches potential decoding errors in order to test each dataclass, until a suitable one (or none is found).  The PR is against master, and when I look at files changed I see (as well as some added warnings and tests):

Screenshot from 2023-09-04 16-00-05

pawelwilczewski:
@idbentley See what the final PR [looked like](https://github.com/lidatong/dataclasses-json/pull/464/files). As per @george-zubrienko suggestion, I refactored the file a bit before finalising the PR.

idbentley:
Oh, I see!  Sorry, I misunderstood your comment @george-zubrienko !   I've updated it to match recommended style!

idbentley:
Hi @george-zubrienko . Would love to see this PR get a review as it will enable a feature to be completed on my team.

I updated the style (as you asked) last week, and and rebased from master today.

Thanks for your consideration!

dataclasses_json/core.py
```diff
@@ -314,14 +314,16 @@ def _decode_generic(type_, value, infer_missing):
...
                 for type_option in type_options:
                     if is_dataclass(type_option):
-                        res = _decode_dataclass(type_option, value, infer_missing)
-                        break
-                if res == value:
+                        try:
+                            res = _decode_dataclass(type_option, value, infer_missing)
+                            changed = True
+                            break
+                        except (KeyError, ValueError):
```
george-zubrienko, line 324: It is unclear to me what is the real life case for this, and why do we skip here instead of failing?

dataclasses_json/core.py
```diff
@@ -314,14 +314,16 @@ def _decode_generic(type_, value, infer_missing):
             type_options = _get_type_args(type_)
             res = value  # assume already decoded
             if type(value) is dict and dict not in type_options:
-                # FIXME if all types in the union are dataclasses this
```
george-zubrienko, line 317: I don't think this is resolved by your commit, thus should be kept

dataclasses_json/core.py
```diff
@@ -314,14 +314,16 @@ def _decode_generic(type_, value, infer_missing):
...
-                #  maybe find the best fitting class in that case instead?
+                changed = False
                 for type_option in type_options:
                     if is_dataclass(type_option):
-                        res = _decode_dataclass(type_option, value, infer_missing)
-                        break
-                if res == value:
+                        try:
+                            res = _decode_dataclass(type_option, value, infer_missing)
+                            changed = True
```
george-zubrienko, line 322: The changed part is not necessary here? As I understand you just add try...except?

george-zubrienko (approved):
LGTM, minor comment re exception catch

dataclasses_json/core.py
idbentley: This PR explicitly targets this comment - this is what it is attempting to solve.

Instead of simply accepting the first dataclass in the list of `type_options`, each type is iterated, a decode attempt is made, and if it succeeds, we use that dataclass - otherwise we attempt to decode the next type in the options array.  In this way, we try to find the best fitting class in the `type_options` list - the best fitting match in the union.

dataclasses_json/core.py
idbentley: We only mark `changed` as true when a decode is successful (when no exception is thrown), so that if no decode attempts are successful a warning will be emitted.

dataclasses_json/core.py
idbentley: This except block catches two situations:

1.  When the dataclass in consideration does not contain one (or more) fields that exist in the `value`.
2.  When the type of one (or more) of the fields that exists in the `value` are not unmarshallable into the type of the corresponding field.

If either of these exceptions are thrown, then we know that the dataclass under consideration is not an appropriate choice to decode into - and we continue onto the next type in the `type_options` array.

This code is exercised in the test `test_deserialize_without_discriminator` - where json strings that do not contain `__type` fields are successfully decoded into the correct types.

The real life use case for this is trying to deserialize data that do not have the ___type discriminator value specified.  This could occur when receiving data that was not encoded by the dataclasses_json project, from an HTTP endpoint for example.

dataclasses_json/core.py
george-zubrienko: But still only one can fit, so this can be done in a more elegant manner with for-comp and `if` checking that the type got decoded without errors. Then you won't need this 'changed' flag and overall readability will improve. However, I leave the final decision on whether to (not) do this small improvement to you :)

dataclasses_json/core.py
george-zubrienko: Understood, thanks for explanation!

dataclasses_json/core.py
idbentley: I updated this to check if the value had changed without the unneeded `changed` variable.

I also updated the warning to be more accurate.

I also added a testcase to ensure that the object remains a `dict` when decoding fails.

george-zubrienko (approved):

george-zubrienko:
Pleasant discussion, elaborate reasoning, all-around great PR. Time to merge!

idbentley:
Great working with you @george-zubrienko !  Thanks for all the help.

PR #487, opened 2023-09-25
Fix PEP 0673 before 3.11

### LINKED ISSUE #459: Dataclasses containing variables that reference themselves in list/dict fail to re-create as original type
### Description

When a dataclass contains a reference to a list/dict of itself in a variable type, converting an object to dict and from dict back to json results in a dictionary in the inner self-typed field instead of the self type.

```python


@dataclass_json
@dataclass
class SpecialLinkedList:
    val: ...
    nexts: list['SpecialLinkedList'] = None



my_list = SpecialLinkedList(val=1, nexts=[SpecialLinkedList(val=2)])

print(my_list == SpecialLinkedList.from_dict(my_list.to_dict()))  # False
```

The code snippet above outputs False.
values:
```python
SpecialLinkedList.from_dict(my_list.to_dict()) == SpecialLinkedList(val=1, nexts={'1': {'val': 2, 'nexts': None}})
my_list == SpecialLinkedList(val=1, nexts={'1': SpecialLinkedList(val=2, nexts=None)})
```

3.10

### Environment description

clean project, only dataclass and dataclasses-json.

george-zubrienko:
Fixes #459 

## Working example on 3.10
```
@dataclass_json
@dataclass
class SpecialLinkedList:
    val: int
    nexts: list['SpecialLinkedList'] = None

my_list = SpecialLinkedList(val=1, nexts=[SpecialLinkedList(val=2)])
print(SpecialLinkedList.from_dict(my_list.to_dict()))
print(my_list)

SpecialLinkedList(val=1, nexts=[SpecialLinkedList(val=2, nexts=None)])
SpecialLinkedList(val=1, nexts=[SpecialLinkedList(val=2, nexts=None)])
```

## Approach
Since before 3.10 we get the type as a plain string, and it is not a fully-qualified import reference, the only option I see is to go over all imported modules and try find the type string there. If that fails, we admit the failure and print a warning to the use saying expected outcome will not the as expected. This code only runs for versions <3.11

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -1,5 +1,6 @@
 import copy
 import json
+import sys
 import warnings
 from collections import defaultdict, namedtuple
 from dataclasses import (MISSING,
@@ -10,7 +11,7 @@ from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
 from typing import (Any, Collection, Mapping, Union, get_type_hints,
-                    Tuple, TypeVar)
+                    Tuple, TypeVar, Type)
 from uuid import UUID
 
 from typing_inspect import is_union_type  # type: ignore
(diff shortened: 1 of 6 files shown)


george-zubrienko:
Will finalize this week - sorry for all the waiting, tough times.

matt035343 (approved):
LGTM

Issue #458, opened 2023-08-02
[BUG] Incompatibility with from __future__ import annotations Affects Overriding of Global Config in Python 3.11


estyxx:
### Description

I found a bug where the use of `from __future__ import annotations` causes `dataclasses_json` to fail when trying to override the global configuration for specific types. 
Since `from __future__ import annotations` changes the behaviour of annotations to be stored as string literals rather than evaluated expressions, the comparison of types in functions like `_user_overrides_or_exts` fails, leading to incorrect matching of types.

This issue particularly affects the overriding of the global config for encoders and decoders, and as a result, the library is not able to match the type correctly when type annotations are imported, leading to a failure in using the correct decoder method.

As a result I'm getting an exception: 
```text
Traceback (most recent call last):
  File "app.py", line 6, in <module>
    p = Page.from_dict(data)
  File ".../lib/python3.11/site-packages/dataclasses_json/api.py", line 70, in from_dict
    return _decode_dataclass(cls, kvs, infer_missing)
  File ".../lib/python3.11/site-packages/dataclasses_json/core.py", line 252, in _decode_dataclass
    init_kwargs[field.name] = _decode_generic(
  File ".../lib/python3.11/site-packages/dataclasses_json/core.py", line 356, in _decode_generic
    res = _support_extended_types(type_arg, value)
  File ".../lib/python3.11/site-packages/dataclasses_json/core.py", line 272, in _support_extended_types
    res = datetime.fromtimestamp(field_value, tz=tz)
TypeError: 'str' object cannot be interpreted as an integer
```

Can be related to #382 

A possible solution could be using `typing.get_type_hints` instead. 

### Code snippet that reproduces the issue
```python
from __future__ import annotations
from dataclasses_json import global_config
from dataclasses_json import LetterCase, dataclass_json
import dataclasses_json
from dataclasses import dataclass
import datetime
from typing import Optional

dataclasses_json.cfg.global_config.encoders[Optional[datetime.datetime]] = datetime.datetime.isoformat
dataclasses_json.cfg.global_config.decoders[Optional[datetime.datetime]] = datetime.datetime.fromisoformat

@dataclass_json(letter_case=LetterCase.CAMEL)
@dataclass
class Page:
    first_published_at: Optional[datetime.datetime]

data = {"first_published_at": "2023-07-24T14:03:57.546769+00:00"}
p = Page.from_dict(data)
print(p)  # Issue: The correct decoder method is not used due to the failure to match the types.
```
[...]

Issue #382, opened 2022-10-19
CatchAll not compatible with __futures__.annotations


stadelmanma:
# Problem
When using `from __future__ import annotations` along side a `CatchAll` field an error `dataclasses_json.undefined.UndefinedParameterError: No field of type dataclasses_json.CatchAll defined` is thrown. I added some print statements to [undefined](https://github.com/lidatong/dataclasses-json/blob/23e5eec42b173cf91a1ff194416ccbc87411df2e/dataclasses_json/undefined.py#L244) and was able to observe with the `__futures__` import the field type comes over as `'CatchAll'` instead of an instance of `typing.Optional[~CatchAllVar]`. 

## Minimal reproducible example:
```
from __future__ import annotations
from dataclasses import dataclass
from dataclasses_json import Undefined, CatchAll
from dataclasses_json import dataclass_json
from datetime import date


@dataclass_json(undefined=Undefined.INCLUDE)
@dataclass
class MyClassWithCatchAll:
    id: int = None
    type: int = None
    created_on: date = None
    details: dict[str, str] = None
    extras: CatchAll = None

MyClassWithCatchAll()

#####
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mstadelman/Development/iml/dataclasses-json/dataclasses_json/undefined.py", line 221, in _catch_all_init
    if _CatchAllUndefinedParameters._get_catch_all_field(
  File "/Users/mstadelman/Development/iml/dataclasses-json/dataclasses_json/undefined.py", line 249, in _get_catch_all_field
    raise UndefinedParameterError(
dataclasses_json.undefined.UndefinedParameterError: No field of type dataclasses_json.CatchAll defined
```

## Potential Fix:
Maybe this is too naive? 
```
    @staticmethod
    def _get_catch_all_field(cls) -> Field:
        catch_all_fields = list(
            filter(lambda f: f.type == Optional[CatchAllVar] or f.type == 'CatchAll', fields(cls)))
        number_of_catch_all_fields = len(catch_all_fields)
        if number_of_catch_all_fields == 0:
            raise UndefinedParameterError(
                "No field of type dataclasses_json.CatchAll defined")
        elif number_of_catch_all_fields > 1:
            raise UndefinedParameterError(
                f"Multiple catch-all fields supplied: "
                f"{number_of_catch_all_fields}.")
        else:
            return catch_all_fields[0]
```

## Versions Checked
Python: 3.9, 3.10
[...]

PR #490, opened 2023-10-04
fix: allow using CatchAll with postponed evaluation of annotations

2ynn:
This PR addresses Issue #382

@george-zubrienko I understand that you had not planned to tackle this issue before v1 and am aware of the discussions in Issue #458 as well.

As mentioned by @USSX-Hares, there's a possibility PEP-649 ends up superseding PEP-563. However, [a decision still has to be made](https://mail.python.org/archives/list/python-dev@python.org/message/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/)

In the meantime, the inability to use `CatchAll` is a major blocker for the full adoption of `dataclasses_json` in my team as we
rely heavily on type hints and have no other option but to postpone their evaluation through `from __future__ import annotations`.

The proposed solution relies on `typing.get_type_hints()` (as suggested by @estyxx) instead of `field.type` (whose type depends on whether or not `from __future__ import annotations` is used)

The globals associated with the decorated class `cls : Type[A]` are gathered using `vars(sys.modules[cls.__module__])`, as recommended in the [PEP563- Resolving Type Hints at Runtime](https://peps.python.org/pep-0563/#resolving-type-hints-at-runtime) section.

### DIFF
--- a/dataclasses_json/undefined.py
+++ b/dataclasses_json/undefined.py
@@ -2,8 +2,9 @@ import abc
 import dataclasses
 import functools
 import inspect
+import sys
 from dataclasses import Field, fields
-from typing import Any, Callable, Dict, Optional, Tuple, Union, Type
+from typing import Any, Callable, Dict, Optional, Tuple, Union, Type, get_type_hints
 from enum import Enum
 
 from marshmallow.exceptions import ValidationError  # type: ignore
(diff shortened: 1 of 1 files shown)


TristanSpeakEasy:
would love to see this get merged, would help us quite a bit

george-zubrienko (approved):
Library has more issues with `from __future__ import annotations` that just CatchAll, but if this fixes the problem for that case specifically and unlocks some, I think it's a good idea to get this merged. Thanks a lot for the contribution!

Issue #472, opened 2023-08-13
_decode_dataclass ignores InitVars because of use of fields


NiroHaim:
### Description

My usecase requires me to work with initvars and override the to_dict method slightly.
Each time I convert my obj to dict, I need to replace a protected attr with its initvar.
The problem is that when I try to create an obj from said dict, I fail because the initvar attr that exists in the class is filtered out.
The initvar is filtered out because in the `_decode_dataclass` function, to create the `init_kwargs` dict, the `fields` function is used, which ignores initvars.

### Code snippet that reproduces the issue

```python
import dataclasses

from dataclasses_json import DataClassJsonMixin


@dataclasses.dataclass
class A(DataClassJsonMixin):
    a_init: dataclasses.InitVar[int]
    _a: int = None

    def __post_init__(self, a_init: int):
        self._a = a_init + 1

    @property
    def a(self):
        return self._a

    @a.setter
    def a(self, value):
        self._a = value

    def to_dict(self, *args, **kwargs):
        base_dict = super().to_dict(*args, **kwargs)
        base_dict['a_init'] = base_dict.pop('_a')

        return base_dict


if __name__ == '__main__':
    A.from_dict(A(a_init=1).to_dict())  # raises: {TypeError}TypeError("A.__init__() missing 1 required positional argument: 'a_init'")
```

### Describe the results you expected

as to_dict is overridable and a_init is an passable attr for A.__init__, I expect a_init to be included in `init_kwargs` in `_decode_dataclass` when creating the obj.

### Python version you are using

python 3.10

### Environment description

dataclasses-json==0.5.14
marshmallow==3.20.1
mypy-extensions==1.0.0
packaging==23.1
typing-inspect==0.9.0
typing_extensions==4.7.1

matt035343:
Seems to be a duplicate of #281

NiroHaim:
Oh! seems I missed it
anyways, I wrote a small function that gets all initvars from a dataclass, only did basic testing of course but maybe this could help:
```python
    @classmethod
    def _get_initvar_names(cls) -> list[str]:
        annotations = cls.__annotations__.copy()

        for base in cls.__bases__:
            if hasattr(base, "__annotations__"):
                annotations.update(base.__annotations__)
        return [name for name, typ in annotations.items() if isinstance(typ, dataclasses.InitVar)]
```
On another note, I saw that the issue remains open from 2021, are there currently any intentions to resolve it?

matt035343:
[...]

Issue #281, opened 2021-03-01
InitVar does not work with dataclasses-json


AbdulRahmanAlHamali:
For example, the following code fails:
```python
@dataclass
class Hello(DataClassJsonMixin):
    a: int
    b: InitVar[int]

Hello.from_dict({"a": 3, "b": 4})
```
With the exception:
```
Traceback (most recent call last):
  File "/home/aalhamali/pycharm-2020.1.3/plugins/python/helpers/pydev/pydevd.py", line 1438, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/home/aalhamali/pycharm-2020.1.3/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/aalhamali/support_channel_scraper/main.py", line 110, in <module>
    Hello.from_dict({"a": 3, "b": 4})
  File "/home/aalhamali/support_channel_scraper/venv/lib/python3.8/site-packages/dataclasses_json/api.py", line 83, in from_dict
    return _decode_dataclass(cls, kvs, infer_missing)
  File "/home/aalhamali/support_channel_scraper/venv/lib/python3.8/site-packages/dataclasses_json/core.py", line 208, in _decode_dataclass
    return cls(**init_kwargs)
TypeError: __init__() missing 1 required positional argument: 'b'
```

After doing some debugging, I see that the library only rotates on the `fields` when it calls the constructor, and `InitVars` are not included in `fields`.

ALiangLiang:
same here

mattany:
@ALiangLiang @AbdulRahmanAlHamali potential workaround for this is using a regular variable and then deleting it in the __post_init__ function

PR #495, opened 2023-10-15
Add support for InitVar

matt035343:
Related to #281 and #472.

Adds support for init_vars by reading internal field metadata and inspecting the init function signature.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -1,4 +1,5 @@
 import copy
+import inspect
 import json
 import sys
 import warnings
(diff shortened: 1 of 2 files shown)


mafrosis:
@matt035343 @george-zubrienko Is it possible to get this rebased and merged? `InitVar` and `dataclasses-json` is currently not possible.

mafrosis:
I did some testing, looks like we may need an extra patch to [handle changing of field names](https://github.com/lidatong/dataclasses-json?tab=readme-ov-file#encode-or-decode-using-a-different-name) during encode/decode.

Something like this [at line 175](https://github.com/lidatong/dataclasses-json/pull/495/files#diff-4ab8dc00de8458e1565cf240d047fbdecd73a5b3a858053c39cf564825f4e531R175):
```python
try:
    field_value = kvs[field_name]
except KeyError:
    # Fields missing from kvs can be skipped
    continue
```

PR #511, opened 2024-01-30
fix: fix deserialization of unions if the type of a field with the same name differs between union members

TristanSpeakEasy:
The type of union that failed is documented in the test I added, but basically if members of a union differed by the type of a field with a matching name, the resultant `AttributeError` wasn't being caught and so deserialization would stop and not all the members of the union would be tried

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -320,7 +320,7 @@ def _decode_generic(type_, value, infer_missing):
                         try:
                             res = _decode_dataclass(type_option, value, infer_missing)
                             break
-                        except (KeyError, ValueError):
+                        except (KeyError, ValueError, AttributeError):
                             continue
                 if res == value:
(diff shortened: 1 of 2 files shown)


TristanSpeakEasy:
@george-zubrienko would be great if this could be merged and a new release made. Otherwise I am staring down the barrel of having to publish our fork on pypi and updating our dependency usage to that as its a blocking issue on our side for customers that have SSE use cases in the AI space

george-zubrienko (approved):
Kinda funky that skipping an exception fixes this, but I can't really see any complications from merging this in, so let's go.

Also, yes, if you need something reviewed asap - please tag in comments, helps me to sort out priorities :)

george-zubrienko:
@TristanSpeakEasy should be available in 0.6.4

TristanSpeakEasy:
Great thank you

PR #524, opened 2024-04-18
fix: apply global config globally

PJCampi:
Use the encoder/decoder registered in the global configuration

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -234,8 +234,7 @@ def _decode_type(type_, value, infer_missing):
         return _decode_dataclass(type_, value, infer_missing)
     if _is_supported_generic(type_):
         return _decode_generic(type_, value, infer_missing)
-    else:
-        return _support_extended_types(type_, value)
+    return _support_extended_types(type_, value)
 
 
 def _support_extended_types(field_type, field_value):
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -227,6 +227,17 @@ def _decode_dataclass(cls, kvs, infer_missing):
     return cls(**init_kwargs)
 
 
+def _decode_type(type_, value, infer_missing):
+    if _has_decoder_in_global_config(type_):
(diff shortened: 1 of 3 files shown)


dataclasses_json/core.py
```diff
@@ -227,6 +227,17 @@ def _decode_dataclass(cls, kvs, infer_missing):
...
 
+def _decode_type(type_, value, infer_missing):
+    if _has_decoder_in_global_config(type_):
+        return _get_decoder_in_global_config(type_)(value)
+    if is_dataclass(type_) or is_dataclass(type_):
+        return _decode_dataclass(type_, value, infer_missing)
+    if _is_supported_generic(type_):
+        return _decode_generic(type_, value, infer_missing)
+    else:
+        return _support_extended_types(type_, value)
```
george-zubrienko, line 238:
```suggestion
    return _support_extended_types(type_, value)
```

george-zubrienko (approved):

george-zubrienko:
Will merge / release by EOW - maybe can squeeze another PR from @PJCampi on top of this one

dataclasses_json/core.py
```diff
@@ -227,6 +227,16 @@ def _decode_dataclass(cls, kvs, infer_missing):
     return cls(**init_kwargs)
 
 
+def _decode_type(type_, value, infer_missing):
+    if _has_decoder_in_global_config(type_):
+        return _get_decoder_in_global_config(type_)(value)
+    if is_dataclass(type_) or is_dataclass(type_):
```
rtsscy, line 233: If this is a refactor from the code in _decode_generic, would this line need to be
` if is_dataclass(type_) or is_dataclass(value)`

PJCampi: Yes you are correct. Let me create a new PR to fix this.

rtsscy: But would this fail actually, because the _decode_dataclass function looks like its expecting a dictionary in this line 
`kvs = {decode_names.get(k, k): v for k, v in kvs.items()}`

PR #525, opened 2024-04-18
fix: support generic dataclass

PJCampi:
support generic dataclass

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -24,7 +24,8 @@ from dataclasses_json.utils import (_get_type_cons, _get_type_origin,
                                     _get_type_arg_param,
                                     _get_type_args, _is_counter,
                                     _NO_ARGS,
-                                    _issubclass_safe, _is_tuple)
+                                    _issubclass_safe, _is_tuple,
+                                    _is_generic_dataclass)
 
 Json = Union[dict, list, str, int, float, bool, None]
 
@@ -259,8 +260,9 @@ def _is_supported_generic(type_):
         return False
     not_str = not _issubclass_safe(type_, str)
     is_enum = _issubclass_safe(type_, Enum)
+    is_generic_dataclass = _is_generic_dataclass(type_)
     return (not_str and _is_collection(type_)) or _is_optional(
-        type_) or is_union_type(type_) or is_enum
+        type_) or is_union_type(type_) or is_enum or is_generic_dataclass
 
 
 def _decode_generic(type_, value, infer_missing):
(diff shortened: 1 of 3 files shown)


george-zubrienko (commented):
@matt035343 wdyt about this? Imo this allows you to squeeze lots of problems in as there are no bounds on T, which means it will always be a string/number/bool, what's the point then?

george-zubrienko:
> It is not a perfect solution at all but the library's coverage improves a little. I can add more tests to better highlight what works and what doesn't.

Please do that - and request review again when ready :)

george-zubrienko:
> After my proposed change, whether the decoding works is down to whether T itself is supported by dataclass_json

Shouldn't we throw in case it is not supported?

PJCampi:
Maybe I should describe my change in more details because the name of the PR (and the ticket I opened) is a bit misleading.

It's not about improving the decoding of type variables in a generic dataclass. That feature currently not supported by the library. The library sets the raw json node to a dataclass field of type T regarldess of its bound, constraints or annotation via `__cls_getitem__`. It never fails:
```
    @dataclass_json
    @dataclass
    class Foo(Generic[T]):
        bar: T

    @dataclass_json
    @dataclass
    class Bar:
        value: int

    assert Foo[Bar].from_json('{"bar": {"value": 1}}') == Foo({"value": 1})  # Bar is not decoded correctly  
```

I am trying to align the behaviour of the library when a generic dataclass is used as type annotation to the behavior I just described.

In the case where a generic dataclass (like `Foo`) is used as part of a type annotation, it is wrapped in an generic wrapper (`_GenericAlias` in python 3.11).

Consider the following example:
```
    @dataclass_json
    @dataclass
    class Baz(Generic[T]):
        foo: Foo[T]
```

On master we silently fail to decode the generic dataclass attribute `foo` and the json node is assigned to the attribute:
```
    assert Baz[int].from_json('{"foo": {"bar": 1}}') == Baz({"bar": 1})
    assert Foo[int].from_json('{"bar": 1}') == Foo(1)
```

After my change we extract the concrete dataclass type from the generic wrapper. The type is now decoded as a dataclass with generic type `T`. The decoding of `foo` becomes consistent with the decoding of an instance of Foo:
```
    assert Baz[int].from_json('{"foo": {"bar": 1}}') == Baz(Foo(1))
    assert Foo[int].from_json('{"bar": 1}') == Foo(1)
```

Throwing an exception instead of silently incorrectly setting a json node to an attribute of complex type would be a good idea but it's orthogonal to my change, and it would be a breaking change to the behaviour of the library. 

I added more test cases now. I did not write any test around type bounds or constraints cause that's not supported by the library anyways.

george-zubrienko:
I am convinced - @matt035343 any additional input or we release?

george-zubrienko (approved):

BKAUTO (approved):

matt035343:
> I am convinced - @matt035343 any additional input or we release?

I think it looks good

matt035343 (approved):

george-zubrienko:
To be released Monday :) fyi @PJCampi - and thanks for contributions and thorough explanations!

PR #529, opened 2024-05-17
Warn about buggy type resolution

DavidCain:
This type annotation resolution is a flawed implementation -- ideally a
proper fix can come later, but for now it's worth warning about the
present behavior.

`dataclasses-json` does not behave correctly when using string-style
annotations (a frequent tactic for annotating types that have not yet
been defined at that point in the file).

The core problem comes from the assumption that type annotations in a
module will match `sys.modules()` -- `maybe_resolved` will frequently
resolve to a different value than a typechecker would!

A short example of the error
============================

`example_types.py`:
```python

@dataclass_json
@dataclass
class Config:
    options: list["Option"]

@dataclass_json
@dataclass
class Option:
    label: str
```

Expected behavior
-----------------
`_decode_items` correctly identifies `Option` as coming from
`example_types`:

```python
>>> from example_types import Config
>>> print(Config.from_dict({"options": [{"label": "repro"}]}).options)
[Option(label='repro')]
```

Unexpected behavior
-------------------
`_decode_items()` incorrectly identifies `"Option"` as from a third
party module that is not in scope at time of annotation:

`repro.py`:
```python
import click.parser  # Has `Option`
from example_types import Config

print(Config.from_dict({"options": [{"label": "repro"}]}).options)
```

```bash
$ python3.10 repro.py
[{'label': 'repro'}]
```

Related fix -- truthiness
=========================
The conditional `if maybe_resolved` is meant to handle the case of an
attribute not being found in the module (it could basically be written
as `if maybe_resolved is not None`). However, that doesn't cover the
case of types that are falsy at runtime!

```python
CustomNoneType = None

values: list['CustomNoneType'] = [None, None]
```

We can just use `hasattr()` instead.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -378,9 +378,11 @@ def _decode_items(type_args, xs, infer_missing):
     hence the check of `is_dataclass(vs)`
     """
     def handle_pep0673(pre_0673_hint: str) -> Union[Type, str]:
-        for module in sys.modules:
-            maybe_resolved = getattr(sys.modules[module], type_args, None)
-            if maybe_resolved:
+        for module in sys.modules.values():
+            if hasattr(module, type_args):
+                maybe_resolved = getattr(module, type_args)
+                warnings.warn(f"Assuming hint {pre_0673_hint} resolves to {maybe_resolved} "
+                              "This is not necessarily the value that is in-scope.")
                 return maybe_resolved
 
         warnings.warn(f"Could not resolve self-reference for type {pre_0673_hint}, "
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,0 +1,35 @@
+from dataclasses import dataclass
+import pytest
+from typing import Optional, Set
+
+from dataclasses_json import dataclass_json
+
+
+@dataclass_json
+@dataclass
+class Config:
+    options: list["Option"]
+
+@dataclass_json
+@dataclass
+class Option:
+    label: str
+
+@dataclass_json
+@dataclass
+class ConfigWithoutStringOptions:
+    options: list[Option]
+
+
+
+
+class TestWarning:
+    def test_warns_about_nondeterministic_resolution(self):
+        with pytest.warns(UserWarning, match="Assuming hint Option resolves to .*"):
+            config = Config.from_dict({"options": [{"label": "scope"}]})
+        assert config.to_json() == '{"options": [{"label": "scope"}]}'
+
+
+    @pytest.mark.filterwarnings("error")
+    def test_plain_type_hints_resolve_correctly(self):
+        ConfigWithoutStringOptions.from_dict({"options": [{"label": "scope"}]})
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -27,7 +27,10 @@ if sys.version_info >= (3, 9):
 
 
 
-@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires generic types for bug")
+@pytest.mark.skipif(
+    not ((3, 9) < sys.version_info < (3, 11)),
+    reason="syntax only valid on Py3.9, but bug disappears after Python 3.11",
+)
 class TestWarning:
     def test_warns_about_nondeterministic_resolution(self):
         with pytest.warns(UserWarning, match="Assuming hint Option resolves to .*"):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,28 +1,33 @@
 from dataclasses import dataclass
 import pytest
-from typing import Optional, Set
+from typing import Optional, Set, List
 
 from dataclasses_json import dataclass_json
+import sys
 
 
-@dataclass_json
-@dataclass
-class Config:
-    options: list["Option"]
+# This test exists *only* to demonstrate a bug on Python 3.9+
+# It uses syntax that is not valid on earlier versions of Python!
+if sys.version_info >= (3, 9):
+    @dataclass_json
+    @dataclass
+    class Config:
+        options: list["Option"]
 
-@dataclass_json
-@dataclass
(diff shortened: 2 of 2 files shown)


DavidCain:
I expect that https://github.com/lidatong/dataclasses-json/issues/442 may well address this issue with a refactor (which is why I didn't make any effort to change behavior here). However, I at least wanted to document this issue & hopefully help others who stumble on it!

DavidCain:
For anybody else reading this (and using this package on Python 3.7 through 3.10) a good workaround is to use `from __future__ import annotations` -- that way you can use [postponed evaluations](https://peps.python.org/pep-0563/):

```diff
+from __future__ import annotations

 @dataclass_json
 @dataclass
 class Config:
-    options: list["Option"]
+    options: list[Option]
```

george-zubrienko:
> For anybody else reading this (and using this package on Python 3.7 through 3.10) a good workaround is to use `from __future__ import annotations` -- that way you can use [postponed evaluations](https://peps.python.org/pep-0563/):
> 
> 
> 
> ```diff
> 
> +from __future__ import annotations
> 
> 
> 
>  @dataclass_json
> 
>  @dataclass
> 
>  class Config:
> 
> -    options: list["Option"]
> 
> +    options: list[Option]
> 
> ```

No you should not use that import with DCJ as it breaks the type handling in a lot of cases. Please review documentation :)

DavidCain:
> @DavidCain is it possible for you to add some units proving the fix?

@george-zubrienko - I can certainly add unit tests. Just for clarity though, are you suggesting that I:
- test this current approach of maintaining the buggy behavior (but emitting warnings?)
- or change the way string annotations are resolved, fixing the bug

> Also note that after 3.11 this is not super relevant since 'self' types are resolved there w/o string hacks

Yup, agreed that Python 3.11 helps! That's why I recommended `from __future__ import annotations` -- it's basically opting into Python 3.11 postponed evaluations.

Though note that the issue demonstrated here is *not* annotating the type of the class itself, but instead annotating a class that comes later in the file and hasn't been interpreted yet at time of declaration.

DavidCain:
> No you should not use that import with DCJ as it breaks the type handling in a lot of cases. Please review documentation :)

@george-zubrienko , I assume you're referring to the line here -- https://github.com/lidatong/dataclasses-json?tab=readme-ov-file#handle-recursive-dataclasses

I understand that the `__future__` import may not be "sanctioned" but given the choice between whatever bugs may arise from Python 3.11-style typing or "pick the first module which happens to have a variable of that name" that's an easy choice.

In other words, using `'Option'` creates a bug 100% of the time, but I haven't seen bugs from using `annotations`). Finally, it _should_ be _safe_ to use Python's `__future__` imports! To not support future imports is to be unable to support newer Python versions.

george-zubrienko:
> Finally, it should be safe to use Python's __future__ imports! To not support future imports is to be unable to support newer Python versions.

You misunderstand the concept of future imports. Those are things that might not even make it into next release. Using those is always at your own risk and no, no stable and reliable code should rely or support future imports. This is especially important when you are working with type handling which in Python is screwed to the rest of the language with rusty nails

george-zubrienko:
> test this current approach of maintaining the buggy behavior (but emitting warnings?)
or change the way string annotations are resolved, fixing the bug

It is a tough one to fix pre-3.11. You'll have play the string matching game which is tough, I fixed that for one of the cases some months ago, but I feel like its just chasing the mice without addressing the root issue. Unfortunately we have to wait for 3.10 EOL, so a warning should be fine, I'd just like to see smth that covers the code you are adding
If you have a good idea for a proper fix, I would be glad to review and get it in if implemented

DavidCain:
> You misunderstand the concept of future imports. Those are things that might not even make it into next release.

Fair enough! I know that any future import _may_ ultimately never be added to the language ("MandatoryRelease may also be None, meaning that a planned feature got dropped or that it is not yet decided.")

https://docs.python.org/3/library/__future__.html

What I meant to say is that all current possible imports in `__future__` already have been released (or _will_ be released). They should all be safe to use.

```python
>>> from __future__ import annotations
>>> annotations.getMandatoryRelease()
(3, 11, 0, 'alpha', 0)
```

> Using those is always at your own risk and no, no stable and reliable code should rely or support future imports.

Disagree -- `__future__` is a canonical way to support multiple Python versions at once.

> This is especially important when you are working with type handling which in Python is screwed to the rest of the language with rusty nails

Hah no disagreement there, but again really can't stress enough how much "pick the first module that happens to have a variable of this name" is also screwed.

DavidCain:
> It is a tough one to fix pre-3.11. You'll have play the string matching game which is tough, I fixed that for one of the cases some months ago, but I feel like its just chasing the mice without addressing the root issue. Unfortunately we have to wait for 3.10 EOL, so a warning should be fine, I'd just like to see smth that covers the code you are adding If you have a good idea for a proper fix, I would be glad to review and get it in if implemented

Great, I'd be happy to add some test coverage for what I've done here! If I have the time for a proper fix, I can follow up with a proposal.

george-zubrienko:
> pick the first module that happens to have a variable of this name" is also screwed.

We have a running joke in the office that all Python development is hope-driven ;)
If you manage to get a quick test for the warning in, I'll release this one and one more for abstract classes on the weekend.
Also small disclaimer, I plan to remove tests for 3.7 and 3.8 as being required. If you run into issues with 3.7 in future contributions, just assert True/accept failure if ver <=3.9

george-zubrienko (approved):
Awesome! Let's get this merged :)

s4ke:
We just ran into the case where according to this PR buggy behaviour logged out huge chunks of actual data into stdout. While I understand where this comes from, this is a bad idea because this can leak PII data into stdout by accident. Plus: The old behaviour was "just working as intended" for our usecase.

PR #532, opened 2024-05-23
feat: support abstract collections

PJCampi:
Typing with abstract collections rather than concrete implementations typically make for a looser and contract based API.
This PR implements decoding to an abstract collections by selecting an implementation that conforms to the abstraction.

### DIFF
--- a/dataclasses_json/core.py
+++ b/dataclasses_json/core.py
@@ -3,6 +3,8 @@ import json
 import sys
 import warnings
 from collections import defaultdict, namedtuple
+from collections.abc import (Collection as ABCCollection, Mapping as ABCMapping, MutableMapping, MutableSequence,
+                             MutableSet, Sequence, Set)
 from dataclasses import (MISSING,
                          fields,
                          is_dataclass  # type: ignore
@@ -10,6 +12,7 @@ from dataclasses import (MISSING,
 from datetime import datetime, timezone
 from decimal import Decimal
 from enum import Enum
+from types import MappingProxyType
 from typing import (Any, Collection, Mapping, Union, get_type_hints,
                     Tuple, TypeVar, Type)
 from uuid import UUID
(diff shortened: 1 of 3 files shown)


PJCampi:
Hey @george-zubrienko, 
I am wondering whether you would have a few minutes at some point to have a look this tiny PR? It would be super helpful.
Thanks!
Pierre-Jean

george-zubrienko:
Hey @PJCampi! I'll do it in the evening today - thanks a lot for keeping things in check!

dataclasses_json/core.py
```diff
@@ -304,11 +316,14 @@ def _decode_generic(type_, value, infer_missing):
 
         # get the constructor if using corresponding generic type in `typing`
         # otherwise fallback on constructing using type_ itself
-        materialize_type = type_
         try:
             materialize_type = _get_type_cons(type_)
         except (TypeError, AttributeError):
-            pass
+            materialize_type = type_
```
george-zubrienko, line 322: Can we wrap this into a method? Would be cleaner and easier to read

george-zubrienko (approved):
LGTM, minor style comment - will auto-approve once you commit :)

dataclasses_json/core.py
```diff
@@ -304,11 +316,14 @@ def _decode_generic(type_, value, infer_missing):
...
+
+        # map abstract collection to concrete implementation
+        materialize_type = collections_abc_type_to_implementation_type.get(materialize_type, materialize_type)
```
george-zubrienko, line 325: This one could be in the same method as well

dataclasses_json/core.py
PJCampi: good point! I have moved the logic to resolve the collection type to decode to in one method now.

PJCampi: Is it possible to release/publish the change as well?

george-zubrienko: yes I'll do it on the weekend

george-zubrienko (approved):

george-zubrienko:
@PJCampi released in 0.6.7

Issue #369, opened 2022-07-13
exclude is not propagated to marshmallow schema


andersk:
```python
from dataclasses import dataclass, field
from dataclasses_json import DataClassJsonMixin, Exclude, config

@dataclass
class MyClass(DataClassJsonMixin):
    my_field: int = field(metadata=config(exclude=Exclude.ALWAYS))

print(MyClass(0).to_dict())
# {}
print(MyClass.schema().dump(MyClass(0)))
# {'my_field': 0}
```

I expect `{}` for both outputs.

This was mentioned by @whisust in https://github.com/lidatong/dataclasses-json/issues/187#issuecomment-767815878 but deserves its own issue.

fawzi-kyndryl:
facing the same issue. any update?

george-zubrienko:
@andersk would you be willing to submit a PR to fix? :)

aidandj:
@george-zubrienko Took a stab at a PR for this

oliverbl:
Are the maintainers looking into the PR by @artificial-aidan ? Issue still exists in current version (.0.6.7)

PR #541, opened 2024-08-07
Propagate exclude to marshmallow

aidandj:
Potential solution for: https://github.com/lidatong/dataclasses-json/issues/369

* Propagate exclude function as metadata on schema field
* Add a post-dump hook to exclude fields as indicated


aidandj:
Looks like I broke things with 3.7. What's the plan for deprecating 3.7? It's been EOL for a year now.

I will remove the walrus operator for now to fix it.

PR #553, opened 2025-01-10
Support poetry-core 2.0 changes

s-t-e-v-e-n-k:
poetry-core 2.0 has released, and it's a lot more strict about the configuration groups and their contents in pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,16 @@
-[tool.poetry]
+[project]
 name = "dataclasses-json"
 version = "0.0.0"
 description = "Easily serialize dataclasses to and from JSON."
-authors = ["Charles Li <charles.dt.li@gmail.com>"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -33,7 +40,7 @@ build-backend = "poetry_dynamic_versioning.backend"
 [tool.poetry-dynamic-versioning]
 enable = true
 
-[tool.poetry.urls]
+[project.urls]
 changelog = "https://github.com/lidatong/dataclasses-json/releases"
 documentation = "https://lidatong.github.io/dataclasses-json/"
```
FliegendeWurst, line 45: I also had to change this to `Documentation =`.

chenrui333, line 45: also missing `Repository` in this section as well

